PR #1, opened 2025-12-07
Initial build

jgouly:
This builds on macOS, and gets a binary that does nothing!

Needs testing on Windows.

Linux mostly uses case sensitive filenames, so probably won't work right away, since all the #includes use lower case, but the files are uppercase on the file system.


Ben3eeE (commented):
[b2e721b](https://github.com/elmadev/elma-classic/pull/1/commits/b2e721bf69f92cc693696cbb89a2173ab4fa5449) commit message is too long

PR #4, opened 2025-12-08
Cross platform fs methods

Ben3eeE:
Just opening what I have to have somewhere to discuss the changes, not sure if I want to merge this as it is now :bear:

### DIFF
--- a/KOZOS.CPP
+++ b/KOZOS.CPP
@@ -10,11 +10,6 @@ double K_pi = 3.141592;
 double K_pip2 = K_pi * 0.5;
 double K_2pi = K_pi * 2.0;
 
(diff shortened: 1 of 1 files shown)


GauChoob:
Can we have more descriptive hibas

KOZOS.CPP
```diff
@@ -14,93 +16,71 @@ void k_memcpy(void* dest, const void* src, size_t n) {
     memcpy(dest, src, n);
```
Ben3eeE: This k_memcpy seems unused but it was unused before I was here so I left it.

GauChoob:
Please keep the 8 character limit for now

PR #13, opened 2025-12-10
Initial SDL Window

jgouly:
This is my attempt at first steps of SDL. It's similar to sdl_sunl branch, but splits everything into SDL.CPP right away.


CUSTOM.CPP
```diff
@@ -1,6 +1,289 @@
 #include "all.h"
 
-char* kodtobillnev(int billkod);
+// Customize kiirashoz kell:
```
Ben3eeE: > The function was originally in CUSTOM.CPP until it started to include windows-exclusive code (DIK codes)

What do you mean "until it started to include". Was that something we added and then removed? Or was it always there?

jgouly: That was written by sunl, I was assuming it was about Across. It seems neither `kodtobillnev` or `CUSTOM.CPP` exist in Across, so maybe it was earlier versions of Elma.

M_DD.H
```diff
@@ -1,14 +1,22 @@
...
 
 class ddpal {
     friend void lassufizre(pic8* ppic, ddpal* ppal);
+#if defined(_WIN64) || !defined(_WIN32)
+    SDL_Color pal[256];
+#endif
 
   public:
+#if defined(_WIN32) && !defined(_WIN64)
     void* pddpal;
```
Ben3eeE: This is pddpal and the other is pal? How does that work :bear: not sure I understand the intent of this commit clearly so I ask for clarification.

jgouly: Because pddpal is used by `F_WIN/W_DD.CPP` and `pal` by SDL.CPP.

SDL.CPP
```diff
@@ -33,5 +37,67 @@ int sdl_init() {
...
+
+    unsigned char* mut = (unsigned char*)SDLSurfacePaletted->pixels;
+    if (Locky0_alul) {
+        // Load screen upside-down
+        for (int y = 0; y < 480; y++) {
+            Mutbuffer[480 - 1 - y] = mut;
+            mut += SDLSurfacePaletted->w;
+        }
+    } else {
+        // Load screen rightside-up
```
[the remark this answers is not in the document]
Ben3eeE: I would suggest writing that instead of the current comments.

directinput_scancodes.h
```diff
@@ -0,0 +1,148 @@
+
```
GauChoob: Can we switch to more modern dik file used in clientside? See my two step commit in my branch. Imo it can be done in a single commit because it only takes one rename. Then it's the same file on both sides

SDL.CPP
```diff
@@ -33,5 +37,65 @@ int sdl_init() {
...
+        return 1;
+    }
+
     return 0;
 }
+
+static unsigned char* Mutbuffer[480];
+static int Bufflocked = 0;
+
+unsigned char** lockbackbuffer(int unused) {
```
GauChoob: If we remove fizikai in the function, should we clean up the calling function too and just remove the unused param now? So we don't have to go back to it again

SDL.CPP
```diff
@@ -0,0 +1,37 @@
+#include "ALL.H"
+#include <SDL.h>
+
+SDL_Window* SDLWindow;
+SDL_Surface* SDLSurfaceMain;
+
+int sdl_init() {
+    if (SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS)) {
+        printf("[%s]\n", SDL_GetError());
```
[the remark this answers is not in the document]
jgouly: F_WIN/W_V.CPP has a hiba that will write to a file / use MessageBox(), stub.cpp just has a printf() hiba. It should be fine.

EDIT: just realised hiba() only supports char*, not printf style formatting. I'll leave it as printf() for now and we can replace it later. Very unlikely to hit these anyway.

PR #20, opened 2025-12-11
translate: ABC8 public interface

jgouly:


Ben3eeE:
if first commit is `fix(ABC8)` second commit should be `translate(ABC8)` :mehutrip:

ABC8.CPP
```diff
@@ -1,20 +1,12 @@
 #include "all.h"
 
-static int Hibaismeretlenkar = 0;
+static int ErrorOnMissingCodepoint = 0;
```
Ben3eeE: Are we going to keep this casing? Or do we use snake case?

PR #21, opened 2025-12-11
Cleanup unused

GauChoob:
Delete some unused stuff

### DIFF
--- a/ADATOK.CPP
+++ b/ADATOK.CPP
@@ -9,7 +9,7 @@ double Talppontegybeolvadasitav;
 double Ugroturelem;
 double Vegenvaras;
 
-double Drsugar, Drtang, Sr;
+double Dr, Sr;
 
 double Fejsugar;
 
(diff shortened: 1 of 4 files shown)


ADATOK.CPP
```diff
@@ -9,7 +9,7 @@ double Talppontegybeolvadasitav;
 double Ugroturelem;
 double Vegenvaras;
 
-double Drsugar, Drtang, Sr;
+double Dr, Sr;
```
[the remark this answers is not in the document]
GauChoob: It's gonna be translated later, should do change when translated

RECORDER.CPP
```diff
@@ -129,13 +129,13 @@ static void filebair( float d1, float d2, float d3 ) {
     fclose( h );
 }*/
 
-int recorder::recall(motorst* pmot, double dt, hanginfo* phinfo) {
+int recorder::recall(motorst* pmot, double ido, hanginfo* phinfo) {
```
jgouly: isn't 'dt' here like delta time / change in time?

LEJATSZO.CPP
```diff
@@ -732,8 +712,6 @@ static int Elozoshowkep1 = 1, Elozoshowkep2 = 1;
 
 // Termeszetes befejezes eseten 0-t ad vissza:
 long lejatszo_r(char* filenev, int showkepmarad) {
-    Kitoltestmegrak = 3; // kirajz meg 3x kirakja keretet
```
jgouly: This one is a bit more confusing since it's a global, but it does seem like the value of 3 isn't used before we get to kirajzol320()..

RECORDER.CPP
GauChoob: So the var is named delta time, but the var is actual regular time and not delta time (ido means time). So I'm clarifying it isn't dt but just regular t (to be translated again later anyways)

PR #42, opened 2025-12-12
fix(topol): fix reading level description

Ben3eeE:
On Windows opening a file in rt mode strips the \r character macos does not do this and the \r character will still be present so we need to replace any possible \r characters.


GauChoob:
Looks ok. Technically checking \r before \n saves a few machine cycles, but even if not fixed can merge I guess

PR #58, opened 2025-12-14
Sound

jgouly:


GauChoob (commented):
This doesn't work, sound still plays when disabled. This should be in the function callbackhang, not the subfunction startwave which is only related to sound events.

I recommend if (Mute || !State->sound_on) so that disabling sound also clears the sound event buffer (even though it probably should be cleared anyways)

edit: re: this commit: https://github.com/elmadev/elma-classic/pull/58/commits/baff8c177ead46e83d8af74ac05f3d209d3e4f1a

GauChoob:
Otherwise yes we can delete hang and nullahang

PR #59, opened 2025-12-14
chore: add include guard to header files

Ben3eeE:


MAIN.H
```diff
@@ -1,3 +1,6 @@
+#ifndef M_MAIN_H
```
jgouly: extra M_

jgouly (approved):

PR #63, opened 2025-12-16
feat: add support for drawing all pics in the background

jgouly:


ALL.H
```diff
@@ -13,6 +13,8 @@
 #define SCREEN_WIDTH 640
 #define SCREEN_HEIGHT 480
 
+constexpr bool ShowForegroundEnabled = false;
```
Ben3eeE: Didn't we say const is all caps? And I vote to use a name similar to eol.

ECSET.CPP
```diff
@@ -1954,31 +1955,35 @@ void betoltecseteket(void) {
...
-        if (xpos > 100) {
-            hiba("betoltecseteket xpos > 100!");
-        }
-        while (pmd) {
-            if (pmd->tavolsag < 500) {
-                if ((pixelek_t)pmd->pixelek >= 10) {
-                    Pecsetfelso->addbytesor(pmd->pixelek, pmd->tavolsag, xpos,
-                                            xpos + pmd->xsize - 1, i, FAZIS_URESRE);
-                    pmd->pixelek = (unsigned char*)PX_URES;
+    if (!EolSettings->pictures_in_background) {
```
GauChoob: Can we maybe take this loop out and put it into a function?

OPTIONS.CPP
```diff
@@ -134,7 +135,14 @@ void options(void) {
         strcpy(Rubrikak[9 + vantagis], "Customize Controls ...");
         strcpy(Rubrikak_tab[9 + vantagis], "");
 
-        val.bead(10 + vantagis, 1);
+        sprintf(Rubrikak[10 + vantagis], "Pics In Background:");
```
GauChoob: strcpy

PR #64, opened 2025-12-16
Beallit

GauChoob:


physics_move.cpp
```diff
@@ -73,34 +88,18 @@ static int valid_anchor_points_new(vekt2 point1, vekt2 point2, rigidbody* rb, ve
 int Konvex = 0;
 int Konkav = 0;*/
 
-/*int Pnyomva = 0;
-
-rigidbody buf_pk;
-vekt2 buf_F;
-double buf_M, buf_dt;*/
+/*int Pnyomva = 0;*/
```
Ben3eeE: Why was this int not deleted? Also comments above this block?

PR #66, opened 2025-12-17
State

GauChoob:
This one was really heavy, but hopefully it's in better shape now


GauChoob:
Dunno what version we want to be, 1.4 seems boring, 2.0 seems confusing

Issue #22, opened 2025-12-11
Delete Paratlankepsorok


GauChoob:
Useless feature of pic8, possibly to obfuscate data

PR #68, opened 2025-12-18
Pic8 part 1

GauChoob:
get_width()/get_height()/get_row() seem like possibly leftovers from planned inheritance or some alternative implementation of fizkep, but I'm not sure Balazs actually knew how to do inheritance. We could replace these all with direct access to the variables which could simplify some things, I'm not sure about people's opinions?

Header file will be cleaned up in part 2


GauChoob:
Also resolves https://github.com/elmadev/elma-classic/issues/22

jgouly (approved):

PR #69, opened 2025-12-18
Sound cleanup

jgouly:


F_WIN/W_V.CPP
```diff
@@ -1,317 +0,0 @@
...
-    if (DwDXVersion < 0x300) {
-        uzenet("You need DirectDraw 3.0 or later installed to run this game!");
-    }
-
-    mk_init();
-
-    // Ezeket atraktam lejatszoba:
-    // starthanghigh();
-    // initdsound();
-    Mute = 1;
```
GauChoob: It's not strictly necessary since sound doesn't play in the menu, but you could maybe initialize the global Mute as 1

F_WIN/W_V.CPP
```diff
@@ -1,317 +0,0 @@
...
-        // Ha memory szerepel text1-ben:
-        if (access("message.inf", 0) == 0) { // Most mem-et is csak inf-re adja:
-            directhiba("message.inf accessed\n", text1, text2, text3);
-        }
-        directhiba("Sorry, out of memory!", NULL, NULL, NULL);
-    }
-    directhiba(text1, text2, text3, NULL);
-}
-
-int s_random(int maximum) { return rand() % maximum; }
```
GauChoob: Why are we dropping s_random when it's not yet implemented properly?

PR #70, opened 2025-12-18
Golyok

GauChoob:
A surprise three-fer


GOLYOK.H
```diff
@@ -1,26 +1,25 @@
...
+struct ball {
+    vekt2 keyframe_r;
+    vekt2 current_r;
+    vekt2 v;
+    double keyframe_rotation;
+    double current_rotation;
+    double angular_velocity;
     double radius;
     double keyframe_time;
-} ball;
```
Ben3eeE: If I know things correctly this only makes a difference in C and is OK to remove? But maybe Joey knows better.

PR #71, opened 2025-12-18
Warning cleanup

jgouly:
Just a few simple warnings removed

### DIFF
--- a/LEJATSZO.CPP
+++ b/LEJATSZO.CPP
@@ -770,8 +770,8 @@ long lejatszo_r(const char* filenev, int showkepmarad) {
 
         if ((Single && befejezte1) || (!Single && befejezte1 && befejezte2)) {
             // Motor hangjat kikapcsolja:
-            setmotor(1, 1.0, 0.0001);
-            setmotor(0, 1.0, 0.0001);
+            setmotor(1, 1.0, 0);
+            setmotor(0, 1.0, 0);
             stopmotor(1);
             stopmotor(0);
             setsurlodas(0);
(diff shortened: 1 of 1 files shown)


OPTIONS.CPP
```diff
@@ -83,11 +83,11 @@ void options(void) {
         }
 
         // Jatekos valasztas:
-        sprintf(Rubrikak[1 + vantagis], "Player A:");
-        sprintf(Rubrikak_tab[1 + vantagis], State->player1);
+        sprintf(Rubrikak[1 + vantagis], "%s", "Player A:");
```
GauChoob: You could just change this to strcpy for the whole block of sprintf?

LEJATSZO.CPP
```diff
@@ -770,8 +770,8 @@ long lejatszo_r(const char* filenev, int showkepmarad) {
 
         if ((Single && befejezte1) || (!Single && befejezte1 && befejezte2)) {
             // Motor hangjat kikapcsolja:
-            setmotor(1, 1.0, 0.0001);
-            setmotor(0, 1.0, 0.0001);
+            setmotor(1, 1.0, 0);
```
GauChoob: This is kind of funny; Balazs mixed up the second and third params

state.cpp
```diff
@@ -185,13 +185,13 @@ static void write_stats_topten(FILE* h, topten* tten, bool single) {
         char time_text[40];
         centiseconds_to_string(tten->times[i], time_text, true);
         fprintf(h, "    ");
-        fprintf(h, time_text);
+        fprintf(h, "%s", time_text);
```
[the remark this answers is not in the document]
jgouly: Possibly, but then have to call strlen() everywhere.

PR #75, opened 2025-12-19
Resizable window/resolution

GauChoob:
Draft

### DIFF
--- a/ALL.H
+++ b/ALL.H
@@ -10,8 +10,10 @@
 #include <unistd.h>
 #endif
 
-#define SCREEN_WIDTH 640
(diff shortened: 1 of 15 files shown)


GauChoob:
I also had a crash after resizing to very small in internal editor, then entering "Play" menu, but dunno if that's the true cause

GauChoob:
Need to recalculate aspect ratio in editor_canvas

mankeli:
This is pretty awesome, would just need some polish

Ben3eeE:
I will close this for now as there is much needed to get it into merge state. It's interesting though.

PR #88, opened 2025-12-22
Vekt2 translation

jgouly:


jgouly:
Marked ready for review, I will squash all the function renaming, left in case it helped reviewing.

jgouly:
Also didn't include renaming the file, or the class. I'm fine with `vekt2`, sunl had renamed it to `vect2`, but for me they're basically the same.

VEKT2.CPP
```diff
@@ -31,8 +31,6 @@ void vekt2::forgatas(double alfa) {
 
 double abs(vekt2 a) { return gyok(a.x * a.x + a.y * a.y); }
 
-double absnegyzet(vekt2 a) { return a.x * a.x + a.y * a.y; }
```
Ben3eeE: I don't understand the commit description https://github.com/elmadev/elma-classic/pull/88/commits/f27d95de813e0f75c8e7a97f70b1ce1841efd3da

VEKT2.CPP
```diff
@@ -19,103 +19,89 @@ vekt2 vekt2::operator-(vekt2 a) { return vekt2(x - a.x, y - a.y); }
...
+    double b = cos(rotation);
     double xo = x;
     x = b * x - a * y;
     y = a * xo + b * y;
 }
 
-double abs(vekt2 a) { return gyok(a.x * a.x + a.y * a.y); }
-
-double absnegyzet(vekt2 a) { return a.x * a.x + a.y * a.y; }
+double abs(vekt2 a) { return square_root(a.x * a.x + a.y * a.y); }
```
Ben3eeE: Why is this method not length instead of abs? It calculates the length of a vector

VEKT2.CPP
```diff
@@ -170,15 +167,15 @@ static int ketoldalon2(vekt2 v1, vekt2 r2, vekt2 v2) {
         masodik = -1;
     }
     if ((elso == -1 && masodik == 1) || (elso == 1 && masodik == -1)) {
-        return 1;
+        return false;
```
GauChoob: The conditions in ketoldalon2 are opposite from ketoldalon3, so I think you inverted true/false here

(that's why ketoldalon2 was permissive and ketoldalon3 was strict in the old naming scheme).

I know you delete this later but this is wrong I think.

VEKT2.CPP
```diff
@@ -130,19 +121,17 @@ vekt2 circles_intersection(vekt2 r1, vekt2 r2, double l1, double l2) {
...
     return (elso && !masodik) || (!elso && masodik);
 }
 
-// Igaz, ha v1 altal meghatarozott egyenest metszi r2,v2,
-// vagyis ha v2 kezdo es vegpontja egyenes ket oldalan fekszik:
-// Ez mar akkor is metszesnek tekinti, ha kozel vannak egymashoz:
-static bool ketoldalon3(vekt2 v1, vekt2 r2, vekt2 v2) {
+// Returns true if the infinite line `v1` and segment `r2`, `v2` intersect.
+// This is inexact and will return true if the line and segment are close.
+static bool line_segment_intersects_inexact(vekt2 v1, vekt2 r2, vekt2 v2) {
```
GauChoob: This line should have been changed in the subsequent commit and I think this specific commit will fail to compile as the function name isn't changed everywhere

GauChoob:
Otherwise I know there's no cleanup commit per se but some thoughts on cleanup:
1) Should probably change (void) function arguments to (). We could do the entire repo at once in theory I guess but if we're finishing up vekt2 we shouldn't skip it so we don't have to go back to it for cleanup/translation purposes. Once it's done it'll be done.
2) Do we care to add param names in the header file? Imo not 100% necessary in most cases, but maybe at least `    vekt2(double, double);` could be changed to `    vekt2(double x, double y);` or equivalent
3) vect2 (vs vekt2) as discussed in discord.
4) As part of cleanup, we should probably make square_root a static method as it's only externally used in background menu ball physics (sqrt could be used here as it's not bike physics)
5) `double x, y;` in vekt2 definition changed to be onto two lines?
6) Rename VEKT2.CPP/H to vect2.cpp or vekt2.cpp

jgouly:
Didn't do constexpr constructor, it's already getting big chnage.

PR #90, opened 2025-12-23
Main cleanup part 1

GauChoob:
Had a few moments so here's some small changes

Resolves https://github.com/elmadev/elma-classic/issues/24
Resolves https://github.com/elmadev/elma-classic/issues/25

### DIFF
--- a/MAIN.CPP
+++ b/MAIN.CPP
@@ -17,22 +17,6 @@ int sdl_init(void);
 int main() {
     sdl_init();
 
-    initqopen();
-
-    void mk_init(void);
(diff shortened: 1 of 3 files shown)


jgouly:
I think we should drop the last commit and keep main in MAIN

GauChoob:
Entry can be a bit platform dependent, but if you prefer main() could call platform_init or something as sdl_init, thereby keeping main in main

jgouly:
Yeah we can look into that at some point!

jgouly (approved):

PR #100, opened 2025-12-27
ddpal

GauChoob:


platform_sdl.cpp
```diff
@@ -108,6 +108,8 @@ platform_pal::platform_pal(unsigned char* tomb) {
     }
 }
 
+platform_pal::~platform_pal() {}
```
Ben3eeE: Why empty destructor?

GauChoob: In preparation for the subsequent commit. The two commits can be squashed together if you prefer

GauChoob:
Missing pcx_save palette adjustment in last commit

jgouly:
This seems more abstracted away / complicated than I was thinking. Instead of `void* impl` it could just be `void *data`. Also instead of `platform_pal`, `palette` is a better name.

For me abstracting SDL is just about keeping SDL out of most files.

I can try show what I was thinking.

PR #101, opened 2025-12-28
translate options

jgouly:


LEJATSZO.CPP
```diff
@@ -16,6 +16,7 @@ struct viewtimest {
 };
 
 static void initsoundlejatszoban(void) {
+    static int Hangmarmegy = 0;
```
GauChoob: I'm on my phone so can't check properly, but there so many sound init vars. Is Hangmarmegy somehow completely droppable? Or maybe it's better to drop Hangenabled?

options.cpp
```diff
@@ -0,0 +1,215 @@
...
+    menu.add_line("- View Box Toggle", x2, yo + dy * 5);
+
+    menu.add_line("T", x1, yo + dy * 6);
+    menu.add_line("- Time Display Toggle", x2, yo + dy * 6);
+
+    menu.add_line_centered("After you have eaten all the fruits,", 320, yo + dy * 9);
+    menu.add_line_centered("touch the flower!", 320, yo + dy * 10);
+
+    mk_emptychar();
+    while (1) {
```
GauChoob: True?

PR #105, opened 2025-12-29
Koveto translation

jgouly:


KOVETO.CPP
```diff
@@ -62,39 +69,51 @@ static int grass_line_heightmap(gyuru* poly, int v1, int v2, int* x0, int cur, i
...
         if (x < cur) {
             continue;
         }
+
+        // We've reached the end of the heightmap.
         if (x - *x0 >= max_heightmap_length) {
             return cur;
         }
+
+        // y = y1 + (dy/dx) * (x1 - x2)
```
Ben3eeE: Strange comment

Ben3eeE: What I mean is can the code just be rewritten to be clearer instead of a comment?

jgouly: Will have a look, I did think about trying to write a bunch here, about linear interpolation and stuff tooo..

GauChoob:
Hibas are untranslated
Koveto hasn't been translated yet

Issue #16, opened 2025-12-10
Check unused variable deletion in vonalrajz()


jgouly:
See https://github.com/elmadev/elma-classic/commit/09a0b8fe3d16341f91e39590d47ce1cbfd953ac7#diff-49bf457487ca330d436ffad4e6357bea475c3254a6aaea18412be18fc0adc28eL255

@GauChoob thinks that possibly the bug is that should actually be used, investigate it.

PR #112, opened 2026-01-01
Display

GauChoob:


GauChoob:
This reverts the variable deleted in https://github.com/elmadev/elma-classic/issues/16 and fixes the bug

PR #113, opened 2026-01-01
Piclist translation

jgouly:


PICLIST.H
```diff
@@ -5,16 +5,16 @@
...
 
-    enum { Picture = 100, Texture, Mask };
+    enum class Type { Picture = 100, Texture, Mask };
 
     int length;
     char name[(MAX_PICLIST_LENGTH + 10) * 10];
-    int type[MAX_PICLIST_LENGTH + 10];
+    Type type[MAX_PICLIST_LENGTH + 10];
     int distance[MAX_PICLIST_LENGTH + 10];
     int clipping[MAX_PICLIST_LENGTH + 10];
```
GauChoob: I guess we can address clipping enum later in TOPOL where it is defined, during the TOPOL translation

jgouly:
Don't merge

jgouly:
It's ok to merge now. Changed `int get_index(const char* name)` to `int get_index(const char* pic_name)` to avoid a name clash.

PR #116, opened 2026-01-02
Qopen translation

jgouly:


QOPEN.H
```diff
@@ -1,11 +1,10 @@
 #ifndef QOPEN_H
 #define QOPEN_H
 
-FILE* qopen(const char* nev, const char* paramok);
+FILE* qopen(const char* filename, const char* mode);
 void qclose(FILE* h);
 int qseek(FILE* stream, long offset, int whence);
```
GauChoob: long to int?

QOPEN.CPP
```diff
@@ -89,13 +89,13 @@ static int NumHandles = 0;
...
-        internal_error("qopen initqopen nelkul!");
+        internal_error("qopen() called before init_qopen()!");
     }
     if (NumHandles == MAX_HANDLES) {
-        internal_error("qopen-ben Nyitva == MAXNYITVA!");
+        internal_error("NumHandles == MAX_HANDLES!");
     }
     if (strcmp(mode, "rb") != 0 && strcmp(mode, "rt")) {
-        internal_error("qopen-ben sem rb, sem rt!: ", filename);
+        internal_error("qopen() mode is not \"rb\"!: ", filename);
```
GauChoob: This hiba is technically inaccurate both prior to and after "rt" being dropped

I would be okay squashing the hiba translations with internals commit

QOPEN.CPP
```diff
@@ -64,14 +67,14 @@ void init_qopen(void) {
...
+    int size = sizeof(res_file[RES_MAX_FILES]);
+    if (fread(ResFiles, 1, size, h) != size) {
         internal_error("init_qopen() cannot read files!");
     }
     decrypt();
 
-    long magic_number = 0;
-    if (fread(&magic_number, 1, 4, h) != 4) {
+    int magic_number = 0;
+    if (fread(&magic_number, 1, sizeof(magic_number), h) != 4) {
```
GauChoob: Dunno if we want to != 4 or != sizeof. Second one is cleaner I guess even though it can mask mistakes. I'm ambivalent

QOPEN.CPP
```diff
@@ -93,8 +93,8 @@ FILE* qopen(const char* filename, const char* mode) {
     if (NumHandles == MAX_HANDLES) {
         internal_error("NumHandles == MAX_HANDLES!");
     }
-    if (strcmp(mode, "rb") != 0 && strcmp(mode, "rt")) {
-        internal_error("qopen() mode is not \"rb\"!: ", filename);
+    if (strcmp(mode, "rb") != 0 && strcmp(mode, "r") != 0) {
```
GauChoob: commit name could be improved "rt" to "r" instead of deleting "rt"

QOPEN.CPP
GauChoob: There's one more fread that uses hardcoded 4 and 4, dunno if we want to change that as well. Again, game will immediately crash anyways if we unintentionally change the width of this var

QOPEN.CPP
```diff
@@ -25,10 +25,9 @@ static void titkosit(void) {
     if (!Leirotomb) {
         internal_error("titkosit-ban !Leirotomb!");
     }
-    int leiromeret = sizeof(fileleiro);
     for (int i = 0; i < FILESZAM; i++) {
         unsigned char* pc = (unsigned char*)(&Leirotomb[i]);
-        for (int j = 0; j < leiromeret; j++) {
+        for (int j = 0; j < sizeof(fileleiro); j++) {
```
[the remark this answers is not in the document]
jgouly: I just did it before, to avoid translating.

GauChoob:
Also need to rename QOPEN file to qopen (or res if you prefer I guess)

changelog.txt
```diff
@@ -29,3 +29,4 @@
 * noretrac.inf (software backbuffer) is no longer supported
 * fordit.inf (invert volt buttons) is no longer supported
 * nemmem.inf (Across memory check) is no longer supported
+* files.inf (elma.res is unused, files are read from ../files) is no longer supported
```
Ben3eeE: This is extremely hard for me to parse what it's actually saying

PR #125, opened 2026-01-05
Load and save settings from/to settings.json

jgouly:


eol_settings.cpp
```diff
@@ -33,7 +33,6 @@ void to_json(json& j, const eol_settings& s) { j = json{FIELD_LIST}; }
     try {                                                                                          \
         s.name = j.value(#name, s.name);                                                           \
     } catch (json::exception & e) {                                                                \
-        internal_error(e.what());                                                                  \
```
GauChoob: How about menu_dialog warning user, instead of hiba?

GauChoob:
Looks good to me as a base

I think we should have a constexpr DEFAULT_pictures_in_background, DEFAULT_center_camera, etc in a future PR so that we can easily reset to default in the in-game menu (along with the MIN/MAX I previously proposed)

Would be nice to have the enum as a string instead of numbers, but it's okay I guess for now

PR #130, opened 2026-01-06
KOZOS translation

jgouly:


KOZOS.H
```diff
@@ -20,9 +20,9 @@ extern double K_2pi;
 #endif
 
 typedef char finame[20];
```
GauChoob: Could we have a better name than finame perhaps? Though I guess eventually it will just turn into a string in the far future

KOZOS.H
```diff
@@ -5,9 +5,9 @@
 #include <sys/time.h>
 #endif
 
-extern double PI;
-extern double HALF_PI;
-extern double TWO_PI;
+constexpr double PI = 3.141592;
```
GauChoob: This isn't really a trivial cleanup, more of a thought. If these 3 pi vars are constexpr, then HALF_PI and TWO_PI are no longer meaningfully distinct from PI, so we could think about getting rid of the vars HALF_PI and TWO_PI and just keeping PI.

Not sure what you guys think

platform_utils.h
```diff
@@ -5,9 +5,9 @@
 #include <sys/time.h>
 #endif
 
-extern double PI;
-extern double HALF_PI;
-extern double TWO_PI;
+constexpr double PI = 3.141592;
```
jgouly: Yes good idea

PR #133, opened 2026-01-07
Fix last warnings on macOS

jgouly:


GauChoob (approved):
First commit is kind of a band-aid but it's fine until we cleanup ecset

PR #134, opened 2026-01-07
refactor(gyuru): drop get_vertex_count

GauChoob:
I'm going to move this commit into its own PR so we can discuss what to do

Currently the polygon/gyuru class has a public var "vertex_count" and a getter "get_vertex_count()" that at the moment both do the same thing.

There are 22 external references to vertex_count, of which 20 directly access polygon->vertex_count and only 2 access polygon->get_vertex_count(). The polygon class is also such a fundamental type that it is unlikely to be significantly modified. In addition, as long as we have non-encapsulated direct access to the "vertices" var then you cannot even have a meaningful encapsulation with get_vertex_count().

The best thing is to simplify to make all 22 references directly access vertex_count. If we want to later add encapsulation to the polygon/gyuru class, it will be a significant refactoring and we can do it at that point when we need it.

### DIFF
--- a/EDITTOOL.CPP
+++ b/EDITTOOL.CPP
@@ -359,7 +359,7 @@ void t_create_vert_nyomva(int mx, int my) {
 
     if (Pgy) {
         // Most fogunk letenni uj pontot:
-        if (Pgy->get_vertex_count() >= MAX_VERTICES) {
+        if (Pgy->vertex_count >= MAX_VERTICES) {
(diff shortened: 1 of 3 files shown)


Ben3eeE:
Why not do it the other way if it bothers? Make the field private for proper encapsulation.

We don't need to fix every encapsulation issue immediately but doing one field is a step in the right direction.

PR #140, opened 2026-01-09
Add opengl renderer

Ben3eeE:


mankeli:
well, gave it a go. start position of animal farm is ~1260 fps with the software renderer. GL renderer gives ~5300fps at that same position. 640x480 window, sandy bridge, nvidia 1080 with the official driver.

gl_renderer.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+static void setup_PBO(int width, int height) {
+    glGenBuffers(1, &pbo);
+    glBindBuffer(GL_PIXEL_UNPACK_BUFFER, pbo);
+    glBufferData(GL_PIXEL_UNPACK_BUFFER, width * height, nullptr, GL_STREAM_DRAW);
+    glBindBuffer(GL_PIXEL_UNPACK_BUFFER, 0);
+}
+
+int gl_init(SDL_Window* sdlWindow, int width, int height) {
```
jgouly: sdl_window

GauChoob:
I kind of still want commit [6140dbc](https://github.com/elmadev/elma-classic/pull/140/commits/6140dbc4c75d6c6e59abb8830997d1e3cbf012dc) to live in `D_PIC.CPP` instead of SDL.CPP, since the FPS logic isn't related to SDL itself (apart from writing the FPS in the window title)

platform_sdl.cpp
```diff
@@ -26,15 +30,24 @@ void message_box(const char* text) {
 static unsigned char** SurfaceBuffer;
 
 void platform_init() {
+    CurrentRenderer = EolSettings->renderer;
```
jgouly: Do we really need to store this in a global?

Ben3eeE: For now yes because there is no support for switching renderer without restarting Elma. That is viewed as a different problem that can be addressed later 👯 
It's now stored once when you start the game and preserved regardless if the setting changes by using a global that is set in `platform_init`.

PR #152, opened 2026-01-11
hhigh

GauChoob:


GauChoob:
I forgot to rename the file as well, but that's because of my subbranch

hhigh.cpp
```diff
@@ -102,11 +107,10 @@ void sound_engine_init(void) {
 // Which sound the motor is currently generating
 enum MotorState {
     MOTOR_IGNITION = 0,
-    MOTOR_IDLE,
-    MOTOR_IDLE_TO_GAS_TRANSITION,
-    MOTOR_GAS_START,
-    MOTOR_GASSING,
-    A_VISSZAMENET
+    MOTOR_IDLE = 1,
```
jgouly: In general we don't need to assign values to enums, are these saved to a file or anything?

HANGHIGH.H
```diff
@@ -15,30 +15,28 @@ enum {
     WAV_TURA
 };
 
-// Csak program elejen allithato:
-// Ha hamis, semmi directsound fv.-t nem hiv:
+// Main global sound variable
```
Ben3eeE: Useless comment?

hhigh.cpp
```diff
@@ -50,64 +52,73 @@ void sound_engine_init(void) {
...
+    WavBank[WAV_FOOD] = new wav("eves.wav", 0.5);         // eating
+    WavBank[WAV_TURN] = new wav("fordul.wav", 0.3);       // turn
+    WavBank[WAV_RIGHT_VOLT] = new wav("ugras.wav", 0.34); // jump
+    WavBank[WAV_LEFT_VOLT] = WavBank[WAV_RIGHT_VOLT];
+    SoundFriction = new wav("dorzsol.wav", 0.44); // friction
+    SoundFriction->loopol(WAV_FADE_LENGTH);
+
+    double volume = 0.26;
+    SoundMotorIgnition =
+        new wav("harl.wav", volume, 0, HARL_MAX_INDEX_1 + WAV_FADE_LENGTH); // ?Harley
```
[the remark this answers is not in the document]
GauChoob: Supposed to be translations of all the filenames, but meaning of harl.wav is not 100% clear. I could change it to something like "might be reference to Harley-Davidson motorcycles" if it's more clear

HANGHIGH.H
GauChoob: It was useful until I deleted the two other sound-initialized variables :P. But since it's disappearing a few commits later, I will delete it

jgouly (approved):

PR #153, opened 2026-01-11
Fixedpoint Rabbithole

GauChoob:
Danger zone


GauChoob:
Dunno what you think about this. It will allow us to easily use 32-bit and 64-bit fixed point integers to fix some bugs in doboz

GauChoob:
Kind of a big rabbithole though

PR #156, opened 2026-01-11
Szakasz Part 2

GauChoob:


Ben3eeE:
Needs rebase

Ben3eeE:
I don't understand these final commits at all.

GauChoob:
Second commit is optional. Lines are allocated in a linked list where every list element contains 500 line segments. So if you have 100,000 lines, for every single line you want to add, you have to iterate through 2000 elements of a linked list every time you add a new line segment.

If we just keep a reference to the end of the linked list we don't need to iterate the list, except when deleting the memory

Tldr it doesn't impact the physics, it just makes the data structure more efficient

The second thing I change is we allocate 5000 line segments at once instead of 500 lines to make the allocation more efficient and the linked list less wrong

Tldr tldr this could also be a vector but I haven't been modernizing the code per se

PR #163, opened 2026-01-12
Resolution option

### LINKED ISSUE #93: Fix 1.001 rounding error with kepmeret and +/- buttons

jgouly:
TODO:
  - r~~ename ALL? Move SCREEN_ to main.h?~~
  - ~~hib patch, keep for now and open issue? redo with the integer type thing I tried before?~~
  - SCREEN_WIDTH -> ScreenWidth?
  - copy_ground -> draw_ground?
Closes #93

### DIFF
--- a/menu_options.cpp
+++ b/menu_options.cpp
@@ -132,7 +132,11 @@ void menu_options() {
             break;
         }
 
-        nav.setup(13 + flag_tag_opt, true);
+        strcpy(NavEntriesLeft[13 + flag_tag_opt], "Resolution:");
+        sprintf(NavEntriesRight[13 + flag_tag_opt], "%dx%d", EolSettings->screen_width,
+                EolSettings->screen_height);
+
+        nav.setup(14 + flag_tag_opt, true);
 
         choice = nav.navigate();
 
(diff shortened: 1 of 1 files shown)


Ben3eeE:
> rename ALL? Move SCREEN_ to main.h?

I suggest moving it somewhere else. It's strange to have a header only for resolution :P but I'm fine either way.

> hib patch, keep for now and open issue?

Isn't there already https://github.com/elmadev/elma-classic/issues/93 for that?

I would be fine merging the hib patch now, referencing this issue for a better fix later. We don't need to have it perfect before shipping and can do incremental improvements. The hib patch is how it is in eol so it's presumably pretty oke.

ECSET.CPP
```diff
@@ -1443,11 +1443,19 @@ static int Xe1, Xe2;
 unsigned char *Foldsor = NULL, *Egsor = NULL;
 
 static inline void copy_sky(unsigned char* pc, size_t offset, size_t size, bool minimap) {
-    memcpy(pc, Foldsor + offset, size);
+    if (!minimap) {
```
Ben3eeE: As an outside reviewer it's a bit strange to see 56019fd and 6ed2796 add the `minimap` param that is unused and then later see it used in 6a3ec13 

It makes the diff for 6a3ec13 to add the param there but it also makes more sense when reading commits in chronological order to understand the changes. Not saying it must be done but I ask you to consider it.

Ben3eeE (approved):

PR #169, opened 2026-01-14
Drop fixedpoint math in affine_pic rendering

GauChoob:


jgouly (changes requested):
Agreed to do this one after translation, it's quite an important change!

PR #187, opened 2026-01-19
Remove unused includes reported by clang-tidy

Ben3eeE:
First commit is maybe controversial because it was added to platform_utils to not have this block everywhere.
But this fixes what clang-tidy reports.

### DIFF
--- a/JATEKOS.CPP
+++ b/JATEKOS.CPP
@@ -1,10 +1,11 @@
 #include "JATEKOS.H"
+#include "keys.h"
 #include "main.h"
 #include "MAINMENU.H"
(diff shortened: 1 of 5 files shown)


GauChoob:
I would prefer that we replace all uses of access(), all at once, with filename_exists(), also changing function access_level_file() to level_file_exists() with same logic

It's not my preference, but I would also be okay to just merge without doing this and creating an issue to replace access everywhere instead

Ben3eeE:
I only wanted to fix the ones that were introduced by ALL.H removal. Other fixes can go into an issue for later imo.

PR #188, opened 2026-01-19
Zoom

jgouly:
TODO: object size stuff

### DIFF
--- a/eol_settings.cpp
+++ b/eol_settings.cpp
@@ -42,6 +42,19 @@ void eol_settings::set_zoom_textures(bool zoom_textures) {
     invalidate_lgr_cache();
 }
 
+/*
+ * This uses the nlohmann json library to (de)serialise `eol_settings` to json.
+ *
+ * from_json() / to_json() can be overloaded to provide custom (de)serialisation for types.
+ *
+ * `FIELD_LIST` is a list of all the fields from `eol_settings` to be put into the json.
+ * `JSON_FIELD` for a public field where all values are allowed, or `JSON_FIELD_PRIV` for a private
+ * field that has some constraints. These macros are used to avoid repeating code.
+ *
+ * The value for a missing field when reading the json is the default value set by the
+ * `eol_settings` constructor.
+ */
+
 void to_json(json& j, const MapAlignment& m) {
     switch (m) {
     case MapAlignment::None:


pic8.cpp
```diff
@@ -80,6 +80,17 @@ pic8::pic8(const char* filename, FILE* h) {
     internal_error("pic8 could not find file extension: ", filename);
 }
 
+pic8* pic8::scale(pic8* src, double scale) {
```
GauChoob: Does not handle cases where destination width or height rounds to 0 (leading to crash)

```
    if (target_width < 1) {
        target_width = 1;
    }
    ```

File: ECSET.CPP [hunk 2]
```diff
@@ -459,6 +459,7 @@ void ecset::mutatotlanit(void) {
 
 // Ptop kerek-jeinek egesz koordinatait kitolti:
 void ecset::kitoltfoodkoordokat(void) {
+    const double offset = (double)(ANIM_WIDTH / 2) * EolSettings->zoom();
```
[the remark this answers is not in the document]
GauChoob: Would `ANIM_WIDTH / 2.0 * EolSettings->zoom()` be more technically correct and also easier to read? (granted that ANIM_WIDTH is an even number so it doesn't impact anything

LGRFILE.CPP
```diff
@@ -660,13 +661,13 @@ lgrfile::lgrfile(const char* lgrnev) {
 
         // Objektumok:
         if (strcmpi(nev, "qkiller.pcx") == 0) {
-            pkiller = new anim(ppic, "qkiller.pcx");
+            pkiller = new anim(ppic, "qkiller.pcx", EolSettings->zoom());
```
GauChoob: Should we just do double zoom = zoom() at the start of this function? I don't have a super strong opinion, but the code gets slightly more complex when we add in LGR13 support so it could be nicer for this var to be shorter

eol_settings.cpp
```diff
@@ -118,8 +122,16 @@ void to_json(json& j, const eol_settings& s) { j = json{FIELD_LIST}; }
     } catch (const char* e) {                                                                      \
         external_error("Invalid parameter in " SETTINGS_JSON "!", e);                              \
     }
+
+#define JSON_FIELD_PRIV(name)                                                                      \
```
GauChoob: eol_settings is getting really hard to read. I understand what it all does, but only if I stare at it for a few minutes. Can you add some comments explaining each part?

GauChoob (commented):
I didn't test the code - I just read through the commits

ECSET.CPP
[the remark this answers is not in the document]
jgouly: Not sure why that would be a pain? Don't need to go around converting code, but can use it where it makes sense!

anim.cpp
```diff
@@ -3,7 +3,7 @@
 #include "KIRAJZOL.H"
 #include "pic8.h"
 
-anim::anim(pic8* source_sheet, const char* error_filename) {
+anim::anim(pic8* source_sheet, const char* error_filename, double scale) {
```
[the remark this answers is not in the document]
jgouly: Doing the comment below makes the code shorter.

eol_settings.cpp
```diff
@@ -20,9 +23,28 @@ eol_settings::eol_settings() {
...
+    zoom_textures_ = false;
     renderer = RendererType::Software;
 }
 
+void eol_settings::set_zoom(double z) {
+    if (z != zoom_) {
+        zoom_ = (z < MIN_ZOOM) ? MIN_ZOOM : (z > MAX_ZOOM ? MAX_ZOOM : z);
+
+        set_zoom_factor();
+        invalidate_lgr_cache();
```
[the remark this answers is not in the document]
GauChoob: Imo better to remove invalidate_ptop here, and just leave it to invalidate_lgr, since the level is technically still valid (later on invalidate_lgr could call invalidate_canvas instead of invalidate_ptop)

anim.cpp
```diff
@@ -28,9 +28,11 @@ anim::anim(pic8* source_sheet, const char* error_filename) {
     // Split the source picture into individual frames
     for (int i = 0; i < frame_count; i++) {
         frames[i] = new pic8(ANIM_WIDTH, ANIM_WIDTH);
+        unsigned char transparent = source_sheet->gpixel(0, 0);
```
GauChoob: "transparency" is what we use elsewhere in the code

eol_settings.cpp
```diff
@@ -20,9 +22,26 @@ eol_settings::eol_settings() {
...
+    }
+}
+
+void eol_settings::set_zoom_textures(bool zoom) {
```
GauChoob: Unclear function param name

eol_settings.cpp
```diff
@@ -115,8 +119,20 @@ void to_json(json& j, const eol_settings& s) { j = json{FIELD_LIST}; }
     } catch (const char* e) {                                                                      \
         external_error("Invalid parameter in " SETTINGS_JSON "!", e);                              \
     }
+
+#define JSON_FIELD_PRIV(name)                                                                      \
```
GauChoob: Imo all fields should be private, because all fields might have unintended side-effects

KIRAJ320.CPP
```diff
@@ -624,8 +624,8 @@ static void kirakegyjatekost(int ajatekos, pic8* ppic, double t, motorst* pmot,
     // Objektumok kirajzolasa, de view-ba csak kesobb kerulnek:
     int balalsox, balalsoy;
     Pecsetalso->getbalalso_int(sarok, &balalsox, &balalsoy);
-    int objminx = balalsox - ANIM_WIDTH - 2;
-    int objminy = balalsoy - ANIM_WIDTH - 2;
+    int objminx = (balalsox - (ANIM_WIDTH * EolSettings->zoom()) - 2);
```
GauChoob: Imo these brackets feel superfluous a bit

I also think I'm going to get a compile-time warning about implicit double->int cast

Maybe `balalsox - (int)(ANIM_WIDTH * EolSettings->zoom()) - 2` or something similar?

GauChoob (approved):
Gonna approve subject to random small nitpicks so we can get this merged

Please see my thoughts about eol_settings (which can be addressed in a future PR). But imo we should think about refactoring eol_settings before we add more settings

PR #241, opened 2026-01-28
Shirts

jgouly:


lgr.cpp
```diff
@@ -426,6 +426,68 @@ static unsigned char* create_timer_palette_map(unsigned char* pal) {
...
+    constexpr int SHIRT_BMP_HEIGHT = 101;
+    constexpr int SHIRT_BMP_ARRAY_SIZE = SHIRT_BMP_WIDTH_PADDED * SHIRT_BMP_HEIGHT;
+
+    FILE* f = fopen(filename, "rb");
+    if (!f) {
+        return nullptr;
+    }
+
+    short magic;
+    if (fread(&magic, 1, sizeof(magic), f) != sizeof(magic) || magic != BMP_MAGIC) {
```
GauChoob: What I've been doing for fread and fwrite for files with binary formats is 

`fread(&magic, 1, sizeof(magic), f) != 2`

That way we make sure that magic is the right type as well as using the hardcoded size of a file format.

I don't know if that makes sense to you and we should continue this pattern

KIRAJ320.CPP
```diff
@@ -582,7 +582,8 @@ static void kibike(int ajatekos, pic8* ppic, double t, vect2 sarok, motorst* pmo
     kidoboz(p_labfej, p_terd, ppic, 0.21, pmk->leg, 0.03, 0.03, pmot->flipped_bike);
 
     // Vezeto:
-    kidoboz(p_csipo, p_valltest, ppic, 0.2, pmk->body, 0.1, 0.05, pmot->flipped_bike);
+    kidoboz(p_csipo, p_valltest, ppic, 0.2, shirt ? shirt : pmk->body, 0.1, 0.05,
```
GauChoob: If this is ending up as two lines, maybe better to extract the shirt ? shirt : body into its own line? Not a very strong opinion

pic8.cpp
```diff
@@ -518,6 +518,75 @@ bool pic8::pcx_save(const char* filename, unsigned char* pal) {
...
+
+    // BMP rows are padded.
+    int padded_width = width;
+    if (width % 4) {
+        padded_width += 4 - width % 4;
+    }
+
+    unsigned char* pixels = new unsigned char[padded_width * height];
+    fseek(f, offset, SEEK_SET);
+    if (fread(pixels, 1, padded_width * height, f) != padded_width * height) {
```
GauChoob: A thought (feel free to decline to do this) - possible to fread directly into pic->get_row(h)?

pic rows are padded to 4 as well, but we could also skip the padded pixels since we don't want to rely on this assumption

lgr.h
```diff
@@ -64,6 +64,7 @@ struct bike_pics {
     affine_pic* forarm;
     affine_pic* up_arm;
     affine_pic* head;
+    affine_pic* shirt;
```
GauChoob: Maybe it's better to already have this separate from bike_pics? Because right now we have a memory leak where the function doesn't delete the shirt when we call delete_bike_pics()

We could delete it in delete_bike_pics(), but imo then we would want to remove it later...

pic8.cpp
jgouly: Then it's a syscall per row, and if we remove the pic8 row padding it's two per row (fread + fseek). I think it's better to just read and memcpy.

KIRAJ320.CPP
```diff
@@ -57,6 +58,22 @@ double Hatarx1, Hatarx2, Hatary1, Hatary2;
 static int Viewxsize, Viewysize;
 static int Viewxorig, Viewxtolas;
 
+affine_pic* shirt = nullptr;
+
+void init_shirt() {
+    constexpr int SHIRT_BMP_WIDTH = 149;
+    constexpr int SHIRT_BMP_HEIGHT = 101;
+
+    pic8* pic_shirt = pic8::from_bmp("bmp/shirt.bmp");
```
GauChoob: Can this be shirt folder instead?

GauChoob (approved):
Approved with minor comments, feel free to merge after evaluating merits of comments

PR #248, opened 2026-01-30
Teljes Part 1/3

GauChoob:


TELJES.CPP
```diff
@@ -47,20 +47,6 @@ void cheatplayer(void) {
     }
 }
 
-/*static void kiraklogot( void ) {
```
Ben3eeE: Why is it deprecated instead of commented out?

PR #250, opened 2026-01-31
Add replay speed controls

Ben3eeE:
This is fast forward, slow motion and pause controls. Rewinding in the next PR because it's more complicated and want to keep the diff small.

### DIFF
--- a/LEJATSZO.CPP
+++ b/LEJATSZO.CPP
@@ -770,10 +770,49 @@ long lejatszo_r(const char* filenev, int showkepmarad) {
     // Motor1->apple_count = Motor2->apple_count = 0;
     long l = 0;
     int jar1 = 1, jar2 = 1;
+
+    double current_replay_time = 0.0;
+    double last_stopwatch = stopwatch();
+    bool paused = false;
+
     while (1) {
         handle_events(); // Billentyut itt olvassuk be
+
+        double now = stopwatch();
+        double dt = (now - last_stopwatch) * 0.0024;
+        last_stopwatch = now;
+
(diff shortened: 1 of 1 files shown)


eol_settings.h
```diff
@@ -110,6 +128,13 @@ struct eol_settings {
     Default<DikScancode> brake_alias_key_player_a_{DIK_UNKNOWN};
     Default<DikScancode> brake_alias_key_player_b_{DIK_UNKNOWN};
     Default<DikScancode> escape_alias_key_{DIK_UNKNOWN};
+
+    Default<DikScancode> replay_fast_2x_key_{DIK_UP};
+    Default<DikScancode> replay_fast_4x_key_{DIK_RIGHT};
+    Default<DikScancode> replay_fast_8x_key_{DIK_PRIOR};
```
GauChoob: Page Up seems like funny key to me. Is that the same default in eolconf? Does joey have opinion?

LEJATSZO.CPP
```diff
@@ -770,10 +770,53 @@ long lejatszo_r(const char* filenev, int showkepmarad) {
...
+            paused = true;
+        } else {
+            paused = false;
+        }
+
+        if (!paused) {
+            current_replay_time += dt * speed;
+        }
+
+        if (current_replay_time < 0.0) {
```
GauChoob: I guess this isn't strictly necessary yet but it's fine

For a future PR would be nice to skip backwards from 0.0 to end of replay somehow

GauChoob (changes requested):

Issue #23, opened 2025-12-11
Delete memtocpy from PLAY


GauChoob:
With associated two konvoda functions and global var

PR #259, opened 2026-02-02
Play

GauChoob:
Fixes https://github.com/elmadev/elma-classic/issues/214
Fixes https://github.com/elmadev/elma-classic/issues/23


Ben3eeE:
Added issue fixes to PR body.

Ben3eeE:
Also isn't this done now so the file should be renamed?

PR #262, opened 2026-02-02
F1 enter spy

Ben3eeE:
This should probably be split up into 3 separate PRs. Basic F1-enter, camera detach and mouse drag support.

Currently there is only one single player disconnect camera button and not a per player thing because it would be so many configuration options to add support for this. We would need different camera move buttons for each player and I don't really see there is much value in this. Meaning it moves the camera for both players in split screen when you detach and is a universal key.

Also the mouse drag is limited to one player, in multiplayer you move both players screens. It wouldn't be too hard to implement multiplayer support I guess, but I also don't have interest in developing it any further than this it works for single player. Those commits can probably be moved to another branch where this stuff is developed if that is a must for this feature, or we can just merge this and improve on it later.

Also I didn't make the function key configurable, we can do it later when there are more function keys with online features.

No changelog message written yet.

### DIFF
--- a/LEJATSZO.CPP
+++ b/LEJATSZO.CPP
@@ -13,6 +13,7 @@
 #include "platform_impl.h"
 #include "timer.h"
 #include "keys.h"
+#include <algorithm>
 #include <cmath>
 #include <cstring>
 #include <directinput/scancodes.h>


LEJATSZO.CPP
```diff
@@ -25,8 +25,51 @@ int Tag = 0;
...
 
+static void clamp_freecam_to_level(motorst* pmot) {
+    double minx, miny, maxx, maxy;
+    // We swap y1/y2 pointers to get level min in maxy and level max in miny
+    Ptop->get_boundaries(&minx, &maxy, &maxx, &miny, false);
+
+    double cam_min_y = -miny;
+    double cam_max_y = -maxy;
+
+    minx = std::min(minx, pmot->bike.r.x);
```
GauChoob: Seems a bit unnecessary to update clamp box with bike position

LEJATSZO.CPP
```diff
@@ -433,14 +480,18 @@ long lejatszo(const char* filenev) {
                 dt = cel - eddig;
             }
 
-            if (!meghalt1) {
+            if (CamMode == CameraMode::None && !meghalt1) {
```
GauChoob: I feel like we should split the lejatszo gameloop physics into two functions:

normal_physics
map_viewer_physics
(eventually) replay_"physics"

Instead of splicing in the cameramode everywhere

Because honestly it should be handled completely differently

And the current implementation is going to make it harder to integrate drive + replay together eventually

```
Maybe extract this section into a function call?

        while (eddig <= cel - 0.000001) {
...
        if (!Single) {
            kulsoresz(Motor2, &State->keys2, &valt2, Rec2, &Viewtime2, eddig, &valt1.showkep,
                      meghalt2);
        }
```

LEJATSZO.H
```diff
@@ -10,9 +10,15 @@ extern int Tag; // Van-e fogocska
...
+// Free camera mode
+enum class CameraMode { None, MapViewer };
+extern CameraMode CamMode;
+extern double CameraX;
+extern double CameraY;
+
 // Azt a nevet kerik, amit topol::topol kap:
 // BestTime beallitasahoz kell nekik:
-long lejatszo(const char* filenev);
+long lejatszo(const char* filenev, CameraMode cameramode = CameraMode::None);
```
GauChoob: Imo there shouldn't be any default, it should always be specified

LEJATSZO.CPP
```diff
@@ -524,6 +544,15 @@ long lejatszo(const char* filenev, CameraMode cameramode) {
                 update_freecam(dt, Motor1);
             }
 
+            if (EolSettings->mouse_drag_camera()) {
+                int mdx, mdy;
+                get_mouse_motion(&mdx, &mdy);
+
+                if (is_left_mouse_down() && (mdx != 0 || mdy != 0)) {
```
GauChoob: Maybe right_click resets CameraOffsetX/Y to 0.0?

GauChoob (commented):
Cool stuff. I didn't test yet ingame but some thoughts about the code

PR #268, opened 2026-02-03
feat(state): save when settings or controls change

Ben3eeE:


menu_controls.cpp
```diff
@@ -449,6 +449,7 @@ void menu_customize_controls() {
         if (choice < 0) {
             eol_settings::sync_controls_from_state(State);
             eol_settings::write_settings();
+            State->save();
```
GauChoob: syncing FROM state doesn't actually update state.dat?

GauChoob: Maybe I read too quickly, will check again later the use of this

PR #272, opened 2026-02-04
Lgr menu option

jgouly:


GauChoob (changes requested):
load_lgr isn't working exactly as expected. You need to try 3 possibilities:

1) Try to load the level's non-default lgr
2) Try to load the override lgr
3) Try to load default

For example, if a level has an lgr of "missing.lgr" and your override is "overcast.lgr", when you open the level, "default" will be loaded instead of "overcast"

menu_options.cpp
```diff
@@ -55,6 +59,49 @@ void menu_help() {
...
+    strcpy(nav.title, "Pick an LGR!");
+
+    finame filename;
+    std::vector<std::string> lgr_names;
+    bool done = find_first("lgr/*.lgr", filename);
+    int count = 0;
+    while (!done) {
+        constexpr int LGR_EXT_LEN = 4;
+        int len = strlen(filename);
+        lgr_names.push_back(std::string(filename, len - LGR_EXT_LEN));
```
Ben3eeE: `emplace_back` :mehutrip:

GauChoob:
Incidentally in fix.txt I accidentally forgot to keep the line Valtozott = 1 and should be MAX_FILENAME_LEN + 4 or something, not MAX_FILENAME_LEN + 40

menu_options.cpp
```diff
@@ -55,6 +59,49 @@ void menu_help() {
...
+        done = find_next(filename);
+        count++;
+        if (count >= NavEntriesLeftMaxLength - 4) {
+            done = true;
+        }
+    }
+
+    find_close();
+
+    std::sort(lgr_names.begin(), lgr_names.end(),
```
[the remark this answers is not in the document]
jgouly: I don't think this is really necessary, will skip it. Most players will have a very small list and likely just pressing 'd' will take them to default, or it'll be like 2 or 3 arrow presses.

lgr.cpp
```diff
@@ -36,13 +36,12 @@ void invalidate_lgr_cache() {
     CurrentLgrName[0] = '\0';
 }
 
-static bool try_access_lgr(const char* lgr_name) {
+static bool try_access_lgr(const char* lgr_name, const char* backup_lgr) {
     char path[30];
     sprintf(path, "lgr/%s.lgr", lgr_name);
     if (access(path, 0) != 0) {
```
GauChoob: Can we use filesystem::exists instead? With possibly an updated function name ("try" is probably not necessary in function name as well)

GauChoob (commented):
Looks very good! I think just a few small details

PR #273, opened 2026-02-05
Transparency.cpp

GauChoob:


GauChoob:
Function description is in .cpp instead of .h file. Forgot to move it over

PR #275, opened 2026-02-05
fix(render): foreground wheel increases in size while turning

### LINKED ISSUE #269: wheel size changes during turn

GauChoob:
fixes https://github.com/elmadev/elma-classic/issues/269

### DIFF
--- a/physics_init.cpp
+++ b/physics_init.cpp
@@ -25,10 +25,6 @@ double HeadRadius;
 
 double ObjectRadius = 0.4;
 
-// Render size of the wheel when not turning, in meters, to avoid the wheel appearing to be inside
-// the ground
-double WheelBackgroundRenderRadius = 0.395;
-
 double LeftWheelDX;
 double LeftWheelDY;
 double RightWheelDX;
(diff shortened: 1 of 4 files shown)


Ben3eeE (approved):

GauChoob:
I thought I had done this, but the var should also be renamed from `WheelBackgroundRenderRadius` to `WheelRenderRadius` (also all caps as a constexpr)

I will let joey do this or I will do this when joey approves

physics_init.cpp
```diff
@@ -25,9 +25,8 @@ double HeadRadius;
 
 double ObjectRadius = 0.4;
 
-// Render size of the wheel when not turning, in meters, to avoid the wheel appearing to be inside
-// the ground
-double WheelBackgroundRenderRadius = 0.395;
+// Render size of the wheel, to avoid the wheel appearing to be inside the ground
+double WHEEL_RENDER_RADIUS = 0.395;
```
Ben3eeE: constexpr in the header instead?

GauChoob: I put it local to the renderer

PR #277, opened 2026-02-05
Ecset

GauChoob:
Dumping my TODO list here so I don't accidentally lose it

```
Two lines:
        int current_x, current_y;
    double width, height;
    int pixel_width, pixel_height;
        vect2 r, v;
        int heightmap_length = 0, x0 = 0;

        canvas init: (for both inits, so twice)
    node_array = node_array_last = NULL;
        width = height = 0.0;
    pixel_width = pixel_height = 0;
    static int ViewLeftBorder, ViewRightBorder;
    unsigned char *DefaultForeground = NULL, *DefaultBackground = NULL;


bool:
    int is_minimap;
int* prev_can_merge
int player1
int line_direction_negative
int again
int first = 1;
int skip = 0;
    int prev_can_merge = 0;
    int done = 0; -> could also be integrated directly into the while loop
canvas::canvas(int minimap)
canvas::render(int player1 + render_minimap
above_is_grass returns bool (condition can be simplified, bug to be documented)
int alternate_distance

parallax as constexpr

        if (x1 < 120 || x1 + width >= pixel_width - 50) {
            return;  // TODO should probably be continue instead of return
        }

fazis/clipping as CanvasClipping

grass functions: remove clipping param

linked_segment_finder from .h to .cpp

#define PixelsDefaultForeground (1)
#define PixelsDefaultBackground (0)
    swap order

#define as constexpr

dest_dist = 1000000; as constexpr
source_dist += 10000; as constexpr (sky)

arrays using CANVAS_MAX_HEIGHT as pointer arrays

pixels_reference as struct or class

struct canvas_linked_segment {
    int width;
    as union -> replace uses in sort_xpos_and_merge, xpos_to_width, draw_segment

NULL -> nullptr
while(1) -> true?

delete friend void segedfv(void); + implementation
delete Osszegszam
delete     void kiegysor_FF(unsigned char* pc, int ye); // fekete-feher

merge kiegysor_A and B

(void) ()

render + render_minimap combined

set_origin -> zoom-independent margins

draw_texture -> vertical out-of-bounds check

draw_grass_stuff
    pre-rendered grass masks
    zoomed grass
    out-of-bounds check
    starting score 2^31 instead of 10000 to avoid high zoom crash
    uphill battle grass bug

render/render_minimap -> out of bounds
render_row -> out of bounds + simplify

create_foreground_grass -> real thickness

canvas::new_node has redundant code
Can also access node_array_index++ directly in the array

Re-order static functions

Bubble sort of nodes (maybe use std::list or forward_list instead? - it seems to use stable sort)

Delete default_ground_to_pointers to support toggling default textures + changeable resolution

texture original_width should use a fixed width of the max screen resolution?

DEBUG: benchmark ecset and lgr loading

draw picture (skip > 60000) use constexpr from lgrfile

simplify consecutive_transparent_pixels/consecutive_solid_pixels

delete     int felso = 1;

simplify
    char texturanev[20];
    strcpy(texturanev, "qgrass");

    // Kikeressuk textura indexet:
    int texture_index = Lgr->get_texture_index(texturanev);

    extract grass distance  int distance = 600; into a global constexpr - maybe double-check it matches with default distance in lgrfile?

delete
#ifdef DEBUG
        internal_error("7yfg4");
#endif

delete
    if (hossz < 1) {
        internal_error("dd387d3e");
    }

simplify Clipping valid_clipping

MetersToPixels / MinimapScaleFactor as own var


        draw_grass_polygons(CanvasClipping_Foreground); -> only for foreground, delete others

QgrassTextureId as pixelek instead of chunk

unnecessary if: x3
    if (CanvasMinimap) {
        delete CanvasMinimap;
    }
    CanvasMinimap = NULL;

create_canvasses as public static function

use std::ceil/std::floor

sort_xpos_and_merge we don't need to restart from the very beginning every time! x 2

// Make sure canvas is unused so far -> extract into function
```

### DIFF
--- a/ECSET.CPP
+++ b/ECSET.CPP
@@ -26,27 +26,6 @@ canvas* CanvasMinimap = NULL;
 
 #define PIXELMASZK (~((uintptr_t)0x7FF))
 
-/*void vizsgal( canvas* pecset ) {
-
-    canvas_chunk_node* futo = pecset->rows_linked[1071];
-    int xtomb[40];
-    int muttomb[40];
-    int kurx = 1;
-    int i = 0;
-    while( futo ) {
-        if( i > 38 )
-            internal_error( "i > 38!" );
-
-        xtomb[i] = kurx;
-        muttomb[i] = (int)futo->pixels;
-
-        kurx += futo->xsize;
-
-        i++;
-        futo = futo->pkov;
-    }
-}*/
-
 static void memory_error(void) {
     external_error("You do not have enough memory to load this level!",
                    "Try to set the graphic detail to low at the options.");


ECSET.H
```diff
@@ -3,127 +3,134 @@
...
-#define PX_EG (0)
-#define PX_URES (2)
-
-struct mdarab {
-    mdarab* pkov;
-    int xsize;
-    int tavolsag;
-    unsigned char* pixelek;
+#define PixelsDefaultForeground (1)
+#define PixelsDefaultBackground (0)
```
jgouly: Shouldn't these be PIXELS_DEFAULT_FOREGROUND? etc

PR #281, opened 2026-02-07
Show last apple time

### LINKED ISSUE #33: Fix ABC8::len()
ABC8::len() has:

```C
       if (!ppsprite[index]) {
            if (index == ' ') {
                if (this == Pmenuabc) {
                    width += SpaceWidthMenu;
                } else {
                    width += SpaceWidth;
                }
                text++;
                continue;
            }
            if (ErrorOnMissingCodepoint) {
                hiba("Missing codepoint in abc8!");
                return 0;
            } else {
                text++;
                continue;
            }
        }
        width += default_spacing + ppsprite[index]->getxsize();
```

ABC8::write()

```C
        int index = (unsigned char)*text;
        if (index == ' ') {
            if (this == Pmenuabc) {
                x += SpaceWidthMenu;
            } else {
                x += SpaceWidth;
            } 
            text++;
            continue;
        }
        if (!ppsprite[index]) {
            if (ErrorOnMissingCodepoint) {
                hiba("Missing codepoint in abc8!");
                return;
            } else {
                text++;
                continue;
            }
        }
```

write() always uses SpaceWidth or SpaceWidthMenu, even if ppsprite has a ' ' codepoint. len() uses the width of the ' ' codepoint if it exists. So there could be a discrepancy if the ABC8 file contains a ' ' codepoint.

jgouly:
Fixes #33

### DIFF
--- a/menu_options.cpp
+++ b/menu_options.cpp
@@ -212,7 +212,11 @@ void menu_options() {
         strcpy(NavEntriesLeft[19 + flag_tag_opt], "Default LGR:");
         strcpy(NavEntriesRight[19 + flag_tag_opt], EolSettings->default_lgr_name().c_str());
 
-        nav.setup(20 + flag_tag_opt, true);
+        strcpy(NavEntriesLeft[20 + flag_tag_opt], "Show Apple Time:");
+        strcpy(NavEntriesRight[20 + flag_tag_opt],
+               EolSettings->show_last_apple_time() ? "Yes" : "No");
+
+        nav.setup(21 + flag_tag_opt, true);
 
         choice = nav.navigate();
 
@@ -352,6 +356,10 @@ void menu_options() {
 
         if (choice == 19) {
             menu_lgr();
+        }
+
+        if (choice == 20) {
+            EolSettings->set_show_last_apple_time(!EolSettings->show_last_apple_time());
         }
 
         if (flag_tag_opt) {


EDITUJ.CPP
```diff
@@ -27,6 +27,7 @@
 level* Ptop = NULL;
 
 abc8 *Pabc1 = NULL, *Pabc2 = NULL;
+abc8* SmallFont = nullptr;
```
GauChoob: SmallFont isn't really part of the editor (Pabc1 and Pabc2 are used exclusively in the editor)

Maybe SmallFont could go into kirajzol since it's only compatible with lgr palette and not menu palette? Or else put in abc8.cpp?

GauChoob (changes requested):
Looks good, a few small comments

physics_init.h
```diff
@@ -58,6 +58,7 @@ struct motorst {
     vect2 body_v;
 
     int apple_count;
+    int last_apple_time;
```
GauChoob: Should this he in valtozok? Stylistic choice I guess but valtozok contains more metadata vs motor data

jgouly: I put it next to apple_count since they're closely related.

Ben3eeE (approved):

abc8.cpp
```diff
@@ -25,26 +34,36 @@ abc8::abc8(const char* filename) {
...
+    std::string path("fonts/");
+    path.append(filename);
+    FILE* h = fopen(path.c_str(), "rb");
+    if (!h) {
+        // If not found, check elma.res
+        h = qopen(filename, "rb");
+        res_file = true;
+    }
     if (!h) {
         internal_error("Could not open abc8 file:: ", filename);
```
GauChoob: Add a comment here to make sure abc8 file is located at fonts/file

LEJATSZO.CPP
```diff
@@ -710,8 +715,11 @@ static bool replaymag(motorst* pmot, player_keys* popciok, valtozok* pvalt, reco
     int objszam;
     while (prec->recall_event(eddig, &wavindex, &hangero, &objszam)) {
         if (objszam >= 0) {
-            int tmpi = 0;
-            spritefeldolgoz(objszam, &tmpi, pmot);
+            int prev_apple_count = pmot->apple_count;
+            spritefeldolgoz(objszam, &pmot->apple_count, pmot);
+            if (prev_apple_count < pmot->apple_count) {
```
GauChoob: != instead of <, anticipating rewind feature in replays? Or maybe just keep it as < until bene adds rewind

GauChoob (approved):

PR #286, opened 2026-02-08
Dik Codes with Modifiers

GauChoob:
Here's a proof of concept

Modifiers are enabled for universal controls at the moment (+/- screen size and screenshot)


GauChoob:
the naming of typedefs DikScancode/short could maybe be improved, but this is a proof of concept
the implementation of is_key_down could maybe be improved, but this is a proof of concept

PR #288, opened 2026-02-09
Debug: tic() toc()

GauChoob:
Measure load time for lgr

Can also be added to ecset/canvas after translation

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -5,22 +5,26 @@ This project uses [Meson](https://mesonbuild.com/) as its build system.
 ### Prerequisites
 
 - **C++ compiler** (GCC, Clang, or MSVC)
+- **Python**
 
 ### Installing Meson and Ninja
 
 **Meson** (1.10.0 or later) and **Ninja** are required to build this project.
 
 - On macOS:
+
   ```
   brew install meson ninja
   ```
 
 - On Linux:
+
   ```
   sudo apt install meson ninja-build
   ```
 
   Or using pip:
+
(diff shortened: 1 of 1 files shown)


Ben3eeE:
The summary says what the error is. `std:format` doesn't exist => Missing header include

jgouly:
Would be better if you could avoid the ifdefs around using these functions.
So I would:
 - add a new flag in meson to enable these checks (checking perf in DEBUG is a bit strange)
 - have START_TIME/END_TIME macros in the header, if this new flag/define is undefined define the start/stop as empty macros

lgr.cpp
```diff
@@ -478,6 +479,10 @@ static unsigned char* create_timer_palette_map(unsigned char* pal) {
 }
 
 lgrfile::lgrfile(const char* lgrname) {
+#ifdef DEBUG
+    start_time(DebugTimer::Lgr);
```
jgouly: another option is to make these objects:
PerfTime Lgr;

Lgr.end(..)

Then no need to update the header when you want a new timer

jgouly: Could also do RAII so all you need is:

PerfTime lgr(std::format("{}.lgr (zoom={:.2f})", lgrname, zoom))

and end() in the destructor (if it wasnt ended by calling end() already, for manual use)

GauChoob:
joey: Quick 4am review, remove the "print loading times for .lgr and .lev files' from meson options and .MD file 
joey: Replace with some generic "enable performance timing checks" or so

PR #295, opened 2026-02-10
Replacement menu_nav Part 3/3

### LINKED ISSUE #258: abc8::has_char and is_char_valid_for_filename usually takes Keycodes as input, but func param is unsigned char
Imo both of these should accept Keycodes or at the very least ints, and reject anything  >= 256

At the moment, passing any Keycode causes an overflow (but luckily the Keycodes don't currently map to any available glyph picture)

@Ben3eeE

GauChoob:
Needs to be tested on real menus


GauChoob:
Ready for review.

This monstrous PR can probably be split into 3 parts:
 * Creating the new menu_nav
 * Migrating to the new menu_nav
 * Deleting the old menu_nav

It's all chained together because the migration to the new class served as the QA for the new class implementation

jgouly:
```
../JATEKOS.CPP:124:59: warning: lambda capture 'eddigjatekos' is not used [-Wunused-lambda-capture]
  124 |             State->players[i].name, NAV_FUNC(&visszater, &eddigjatekos) {
      |
```
Also old code

```
if (A) {
        strcpy(State->player1, State->players[eredmeny - 1].name);
    } else {
        strcpy(State->player2, State->players[eredmeny - 1].name);
    }
```
but it seems new code only has strcpy to player1?

PR #296, opened 2026-02-10
Translate main menu

Ben3eeE:


jgouly (approved):

main.cpp
```diff
@@ -28,6 +29,8 @@ void delay(int milliseconds) {
 eol_settings* EolSettings = nullptr;
 
 int main() {
+    srand((unsigned)clock());
```
GauChoob: Can we delete the 2 other srand() function calls?

Maybe (unsigned int) is better since we are converting from clock_t and not just int to unsigned int

recorder.cpp
```diff
@@ -555,9 +540,9 @@ int recorder::load_rec_file(const char* filename, int demo) {
...
     }
 
-    int level_id = Rec1->load(filename, h, demo, true);
+    int level_id = Rec1->load(filename, h, true);
     if (MultiplayerRec) {
-        Rec2->load(filename, h, demo, false);
+        Rec2->load(filename, h, false);
     }
 
     if (demo) {
```
GauChoob: Since you're doing rec cleanup as well, the param demo should be bool

menu_main.cpp
```diff
@@ -0,0 +1,322 @@
...
+                continue;
+            }
+
+            if (CtrlAltPressed) {
+                int time = Rec1->frame_count();
+                if (MultiplayerRec && Rec2->frame_count() > time) {
+                    time = Rec2->frame_count();
+                }
+
+                time = (int)(time * 3.3333333333333);
```
GauChoob: 100/RECORDER_FRAMERATE

menu_main.cpp
```diff
@@ -0,0 +1,322 @@
...
+                }
+            }
+        }
+        MenuPalette->set();
+    }
+}
+
+static void menu_demo() {
+    int demo_replay_count = 3;
+    char demo_names[14][14] = {"demor1.rec", "demor2.rec", "demor3.rec"};
```
GauChoob: Doesn't need to be an array of length 14 since there's only 3 demo files
Can demo_replay_count be a constexpr based on the array length?

menu_main.cpp
```diff
@@ -0,0 +1,322 @@
...
+    }
+}
+
+static void menu_prompt_exit() {
+    menu_nav val;
+    val.selected_index = 0;
+    val.x_left = 300;
+    val.y_entries = 200;
+    val.dy = 40;
+    val.y_title = 50; // usually this is not set
```
GauChoob: Imo this comment can be deleted

menu_main.cpp
```diff
@@ -0,0 +1,322 @@
...
+                       (pl == two_before_previous_played_rec && count > 3)) {
+                    pl = generate_random_number() % (count - 1);
+                }
+                two_before_previous_played_rec = previous_played_rec;
+                previous_played_rec = pl;
+
+                MenuPalette->set();
+                loading_screen();
+
+                // 0 param means not reading from resource file:
```
GauChoob: Don't think this comment necessary

PR #302, opened 2026-02-11
feat(editor_topology): allow Start object to be in ground

jgouly:
Up for discussion


GauChoob:
```
sunl: I double checked and starting in the ground is allowed by both SLE and ALE
sunl: and elma.online editor :P
sunl: So internal editor is pretty much the only one that disallows it currently
```

PR #305, opened 2026-02-12
Keycode -> char overflow bugs

GauChoob:
Commits are spun out from menu_nav commits


Ben3eeE:
Oh and this diff is not a complete working solution, just something to show the idea.

PR #324, opened 2026-02-13
Editplay

GauChoob:
Do you guys prefer moving this function into EDITUJ (additional dependency: lejatszo.h) or keeping it separate?


jgouly:
Fine by me to move into UJ

PR #328, opened 2026-02-16
Leptet

GauChoob:
Plz don't forget about poor ecset :P

Careful cleanups that shouldn't have any impact on physics

Variables and comments are very verbose because the physics is very complicated

### DIFF
--- a/LEJATSZO.CPP
+++ b/LEJATSZO.CPP
@@ -5,7 +5,7 @@
 #include "fs_utils.h"
 #include "sound_engine.h"
 #include "KIRAJZOL.H"
-#include "LEPTET.H"
+#include "physics_forces.h"
 #include "level.h"
 #include "lgr.h"
 #include "main.h"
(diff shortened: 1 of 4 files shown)


LEJATSZO.CPP
```diff
@@ -161,10 +161,11 @@ static void belsoresz(motorst* pmot, player_keys* popciok, valtozok* pvalt, reco
...
     // LEPTET!!!!!!!:
     // 0-meghalt, 1-megnyerte, 2-semmi kulonos
-    leptet(pmot, eddig, dt, is_key_down(popciok->gas),
-           is_key_down(popciok->brake) || is_key_down(popciok->brake_alias), ugrik1, ugrik2);
+    simulate_bike_physics(pmot, eddig, dt, is_key_down(popciok->gas),
+                          is_key_down(popciok->brake) || is_key_down(popciok->brake_alias), ugrik1,
+                          ugrik2);
 
-    int eredmeny = vizsgalat(pmot);
+    int eredmeny = collect_objects_and_head_death(pmot);
```
Ben3eeE: I don't like this name. Just `check_object_collision` would be descriptive enough.

PR #329, opened 2026-02-16
Refactor input

Ben3eeE:
This implements the idea from https://github.com/elmadev/elma-classic/pull/305#issuecomment-3904784379

Separates the input system into two distinct paths:
  - Text input: KeyBuffer now only contains actual typed characters (from SDL_TEXTINPUT), accessed via has_text_input() / pop_text_input()
  - Control keys: checked directly via scancode functions (was_key_just_pressed(), was_key_pressed_or_repeated()) instead of going through the KeyBuffer

This eliminates the Keycode enum and the Keycode-to-char overflow that caused #258 there's no longer a code path where KEY_UP or KEY_PGDOWN gets stuffed into an unsigned char.

Other changes:
  - Numpad keys (arrows, enter, pgup/pgdn, delete) are mapped to their navigation equivalents in the platform layer based on NumLock state, so callers just check DIK_UP / DIK_RETURN etc. without needing || DIK_NUMPAD*
  - SDLToKeycode mapping table removed (no longer needed)
  - left/right_mouse_clicked() replaced with was_left/right_mouse_just_clicked() (per-frame, no hidden handle_events() call)
  - menu_pic::loop() extracted to reduce boilerplate in simple menu screens

Needs some testing, especially num lock behavior on Windows because I don't have a num lock key. I am just guessing :mehutrip: but this changes behavior a bit from EOL. In EOL num lock is not respected and numpad is always treated with movement key behavior. Doing both text input and movement conflicts with SDL textinput events.

Closes https://github.com/elmadev/elma-classic/pull/305

### DIFF
--- a/platform_impl.h
+++ b/platform_impl.h
@@ -27,8 +27,6 @@ void unlock_frontbuffer();
 
 void get_mouse_position(int* x, int* y);
 void set_mouse_position(int x, int y);
-bool left_mouse_clicked();
-bool right_mouse_clicked();
 bool was_left_mouse_just_clicked();
 bool was_right_mouse_just_clicked();
 void show_cursor();
--- a/platform_sdl.cpp
+++ b/platform_sdl.cpp
@@ -15,8 +15,6 @@ static SDL_Window* SDLWindow = nullptr;
 static SDL_Surface* SDLSurfaceMain = nullptr;
 static SDL_Surface* SDLSurfacePaletted = nullptr;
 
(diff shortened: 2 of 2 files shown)


GauChoob:
I had taken a quick look at this when looking at part 1, and overall everything seemed to be fine, though I would like to merge menu_nav instead of needing to update menu_nav_old (not 100% necessary but would be cleaner)

Will take a quick look when it is rebased

PR #330, opened 2026-02-16
Input split

Ben3eeE:


best_times.cpp
```diff
@@ -105,16 +106,8 @@ static void render_topten(topten* tten, const char* header, int single) {
...
-        if (has_keypress()) {
-            Keycode c = get_keypress();
-            if (c == KEY_ESC || c == KEY_ENTER) {
-                return;
-            }
-        }
-        men.render();
-    }
+    men.loop(
+        [&] { return !(was_key_just_pressed(DIK_ESCAPE) || was_key_just_pressed(DIK_RETURN)); });
```
GauChoob: Why are we capturing the scope `&`?

GauChoob (approved):
I don't agree with DialogResult since it just makes everything more complicated. menu.loop() and menu_dialog() should just return the pressed Dik Scancode and in the rare cases that we need it we can check it.

But otherwise the PR seems really good

PR #334, opened 2026-02-17
Gravarrow bmp

Ben3eeE:
I am not making more edits to this. Take it or leave it.

### DIFF
--- a/KIRAJ320.CPP
+++ b/KIRAJ320.CPP
@@ -17,6 +17,7 @@
 #include "pic8.h"
 #include "platform_utils.h"
 #include "physics_init.h"
+#include "renderer/object_overlay.h"
 #include "timer.h"
 #include <algorithm>
 #include <cmath>
@@ -83,6 +84,7 @@ static void init_shirt() {
 
 void init_renderer() {
     init_shirt();
+    init_gravity_arrows();
 
     SmallFont = new abc8("small.abc");
     SmallFont->set_spacing(1);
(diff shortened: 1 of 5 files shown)


KIRAJ320.CPP
```diff
@@ -32,6 +32,53 @@ int Kitoltestmegrakkezd = 10;
...
+    }
+}
+
+static void draw_gravity_arrow(pic8* ppic, int cx, int cy, object::Property property) {
+    pic8* arrow = GravityArrows[(int)property];
+    if (!arrow) {
+        return;
+    }
+
+    int half = (int)(ANIM_WIDTH * EolSettings->zoom()) / 2;
```
GauChoob: Imo it's better to just resize the bmp to 40x40 instead of doing all this calculation to calculate the centerpoint.

If you don't want to support resizing the bmp in this PR (which is honestly just a one-liner so imo you should do it), then just hiba if the bmp dimensions are not ANIM_WIDTH x ANIM_WIDTH.

I don't think it is realistic to want gravity arrows that are larger than the anim image itself, so we don't need to support this centerpoint calculate stuff

KIRAJ320.CPP
```diff
@@ -721,6 +770,12 @@ static void kirakegyjatekost(int ajatekos, pic8* ppic, double t, motorst* pmot,
         }
 
         blit8(ppic, pobjpic, pker->canvas_x - balalsox, pker->canvas_y - balalsoy + dy);
+
+        if (EolSettings->show_gravity_arrows() && pker->type == object::Type::Food &&
+            pker->property != object::Property::None) {
+            draw_gravity_arrow(ppic, pker->canvas_x - balalsox, pker->canvas_y - balalsoy + dy,
```
GauChoob: I don't agree that the gravity arrow should move up/down with the phase (dy). Specifically, the future circular object overlay would not float up/down, and neither should the gravity arrow overlay

GauChoob: Setting then? Should be simple bool

PR #335, opened 2026-02-18
Replacement menu_nav Part 3/3 B

GauChoob:
I'm having issues with rebasing and stuff showing merge conflicts so I'm copying the stuff to a new branch to try


GauChoob:
```
https://github.com/elmadev/elma-classic/commit/0ef55b6307f94b1bdf2037adb3395cb30321fa23

MenuLevel::Null is ugly

```

>>>I agree, I am open to alternative solutions.

```

https://github.com/elmadev/elma-classic/commit/be392fbe0d06c64025ad54a8b4bec1edcbc6e282

I feel like this needs to be split up to review, it's changing a lot of stuff that seem unrelated to menu nav itself. Like adding new internal errors and changing dik_to_string to std::string. It's messy to read as it is. Hard to review.

```

>>>Agree. This PR can probably be split into each file and then I can split this commit into multiple parts. I might get around to doing this if joey get to it first

```

https://github.com/elmadev/elma-classic/commit/0ff69c0b3b8c4fd998d09793c0d6abc78cdeda2f

This should just change to not leak the menu dialog result from load_replay

```

>>>Can you clarify what you mean or suggest a change? Unclear what you are trying to say. If you are asking why load_replay returns a DikScancode, it's to prevent an infinite unescapable loop in Randomizer when 100% of your replays are unloadable. That's why Esc causes a return; while other keys cause a continue; in the parent function. I'm open to an alternative suggestion. I'm also open to slightly changing Randomizer behaviour if it makes more sense to people.

```

https://github.com/elmadev/elma-classic/commit/3251489454530968eda6b132b85db36f427e6e25

Doesn't explain if this is a fix from a regression we made or a bug in elma in the commit. I have no idea how to review this. Needs a changelog entry if it's a bug in elma. Also I don't see how this is related to menu refactoring.

```

>>> This is not a regression - it's just unintuitive behaviour (feels weird just like escing from internal level in across). The reason why the change is made is two reasons:
1) The behaviour is weird and it really should be a break, not a return
2) When we refactor into subfunctions and subsubfunctions in the following commits, it's difficult to return up two function levels in a row. Instead of coming up with some wonky way of returning from two function levels (like throwing an exception or checking extra parameters), we can fluidly use the fix to only need to return 1 single level up, keeping all the code simple and the behaviour more intuitive.
Sure, we can changelog (some very minor things have escaped the changelog but I don't mind adding an entry for this)

```

https://github.com/elmadev/elma-classic/commit/022c0242064f71151c99324501f64595827bbd66

Passes std::string without reference, not in a hot code path though. No idea what is meant by "show true destination filepath" at all. No idea what it "simplifies". Also I don't see how this is related to menu refactoring.

```

>>> I passed it without reference on purpose, because I merged replay_filename and safe_name into the same variable, so the new string isn't const anymore. As for why I did this refactor at all, there are two reasons:
1) We don't need to calculate and maintain the same identical destination filepath in two different functions: 1) setup_render_directory() and 2) menu_replay() or its future subfunction replay_render(). At the moment, setup_render_directory() has extra transformations on the filename so if we do some small changes the filepath as specified by menu_replay() might not be the actual filepath generated by setup_render_directory() - or we might forget to update at both spots in the code.
2) Doing this change pre-emptively simplifies the function replay_render by only needing to use a single unaltered filename

```

https://github.com/elmadev/elma-classic/commit/4cc603c18ac53aa602ff5c7daa1418a855596d5f

Should probably be replay_randomizer

```

>>> Sure

```

Have not reviewed this commit or later: https://github.com/elmadev/elma-classic/commit/ff87e0f1eebb3189e546617a9ae78bca7e428964
```

Ben3eeE (approved):

PR #337, opened 2026-02-18
Input split2

Ben3eeE:


main.cpp
```diff
@@ -79,8 +79,8 @@ static void handle_error(const char* text1, const char* text2, const char* text3
     if (ErrorGraphicsLoaded) {
         render_error(text1, text2, text3, text4);
```
jgouly: I wonder if we can eventually replace render_error() with a menu_dialog, seem to do essentially the same. (not for this PR osv)

JATEKOS.CPP
```diff
@@ -17,41 +19,40 @@ static int bevesznevet(char* nev, int visszater) {
...
-                }
+            menu_exit();
+        }
+        if (was_key_just_pressed(DIK_RETURN)) {
+            if (i > 0) {
+                return 1;
             }
+        }
+        if (was_key_pressed_or_repeated(DIK_BACK)) {
+            // <- (torles):
```
jgouly: Can just drop that comment.

PR #343, opened 2026-02-19
Input split6 (Editor)

Ben3eeE:
Everything is ready but code doesn't make sense until 7 is merged :cow:

Yes this changes the public mouse API of platform. I have no idea how to do it without that because the existing methods are calling `handle_events` Only way to keep the public API is to do everything in a single commit because otherwise mouse stuff wouldn't work properly between commits. The new names are similar to keyboard api so it rolls nicely off the tongue.

Now all the splits have open PRs. I guess menu_nav will be the last because of how it conflicts with the ongoing refactoring so I will update that to add the final removal/cleanup commits once 6 and 7 are merged so that one will be last.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -30,6 +30,7 @@
 * Numpad NumLock state is now respected: navigation when NumLock is off, digit input when on
 * Bike controls configured to numpad keys now work regardless of NumLock state
 * You can now use mouse scroll wheel to navigate the menu
+* Mouse scroll now works in the editor for selecting level, lgr and pictures
 * Elma now saves State.dat when you change options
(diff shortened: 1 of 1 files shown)


Ben3eeE:
Fixed the first commit to have scrolling anywhere (Preserve old behavior with new system as the previous support was global through spamming KEY_UP/KEY_DOWN into the buffer).
Added the boundary check for scrolling in a subsequent commit and also the missing changelog entry.

GauChoob (approved):
I disagree with scroll being in same commit, but at the same time I don't care so I'm approving

PR #344, opened 2026-02-20
Remplacement menu_nav REPLAY

GauChoob:
This is a subsection of Part 3 - only the replay()-related menu_nav commits.

bene recommended changes that were done:

* load_replay returns LoadReplayResult which returns a value stating what to do instead of the key itself
* changelog entry added
* I renamed to replay_randomizer
* I ignored bene's comment about destination filepath

### DIFF
--- a/menu_main.cpp
+++ b/menu_main.cpp
@@ -125,61 +125,45 @@ static void replay_randomizer(std::vector<std::string>& filenames) {
 }
 
 static void menu_replay() {
-    finame filename;
     std::vector<std::string> replay_names;
-    bool done = find_first("rec/*.rec", filename);
 
+    menu_nav nav("Select replay file!");
+    nav.add_row("Randomizer", NAV_FUNC(&replay_names) { replay_randomizer(replay_names); });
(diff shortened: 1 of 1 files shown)


LEJATSZO.CPP
```diff
@@ -963,13 +963,12 @@ long lejatszo_r(const char* filenev, int showkepmarad) {
 }
 
 // Sets the path of VideoOutputDirectory to "./renders/{replay_stem}/"
-static void setup_render_directory(const char* replay_filename) {
-    std::string safe_name(replay_filename);
-    if (safe_name.length() > MAX_FILENAME_LEN + 4) {
-        safe_name.resize(MAX_FILENAME_LEN + 4);
+void setup_render_directory(std::string replay_filename) {
+    if (replay_filename.length() > MAX_FILENAME_LEN + 4) {
```
jgouly: I think we can just drop this code, replay_filename should have already been validated and length doesn't matter for the ouput pcx files. Then replay_filename can be const &

PR #350, opened 2026-02-20
Change resolution without restarting elma #2 Part 1/2

GauChoob:
This is a combination of https://github.com/elmadev/elma-classic/pull/75 and bene's PRs to keep things more simple and also thinking forward to resizable windows and zoom changing

Here are the differences from bene code:

1) Instead of manually managing pic8 subviews (Bpic and Lockbuff), we streamline the subview code (which I was planning on doing in D_PIC.CPP anyways) and also avoid small memory leaks at the same time. It makes the PR slightly longer but the resultant code overall shorter and more clean.
2) No change to bene's compute aspect ratio dynamically
3) Instead of manually setting palette to PaletteMenu at two different places, just remember the current palette (so now surface can be lost anywhere and not only in menu)
4) Instead of deleting and recreating window, making the window disappear for approximately 1 second, use SDL_Resize which instantly resizes the window
5) Change to bene's reinit_menu_pictures(): instead of creating ScreenBuffer immediately, we allow menu_pic to recreate ScreenBuffer. This allows menu_pic to detect when surface is lost and re-render the menu (bene's code: menu would be glitched until menu_nav->navigate() would be called a second time, which happens immediately in options.cpp but doesn't happen in all menus)
6) Change to bene's update resolution: elma resolution update code moved to D_PIC instead of platform_sdl
7) Change to bene's update resolution: instead of calling invalidate_level, the canvas only has a very small dependency on the resolution: lgr's default textures are tiled to the screen width. We drop a small function (that has dubious value in increasing rendering speed) to get rid of the dependency. Now we don't need to reload the entire level if the resolution changes. This also opens up future features like an in-game toggle for default textures ground/sky if we want later on.
8) Addition to bene's update resolution: lgr default textures need to be updated to avoid ASAN issues
9) No change to bene's resolution picker screen

### DIFF
--- a/D_PIC.CPP
+++ b/D_PIC.CPP
@@ -6,7 +6,7 @@
 int SCREEN_WIDTH = 640;
 int SCREEN_HEIGHT = 480;
 
-static pic8* Lockbuff = NULL;
+static pic8 Lockbuff = pic8();
 
 static int Backpiclocked = 0, Frontpiclocked = 0;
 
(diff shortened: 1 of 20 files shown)


GauChoob:
Rebased and added gl_resize under bene's instructions

PR #356, opened 2026-02-20
Somewhat-less-but-still-trivial ecset

GauChoob:
Dumping my TODO list (it's a lot shorter now!)

Mainly linked_list -> std::list and bugfixes left after this

```

linked_segment_finder from .h to .cpp

create_foreground_grass -> real thickness

Re-order static functions

Bubble sort of nodes (maybe use std::list or forward_list instead? - it seems to use stable sort)

texture original_width should use a fixed width of the max screen resolution?

use std::ceil/std::floor

sort_xpos_and_merge we don't need to restart from the very beginning every time! x 2

```

### DIFF
--- a/ECSET.CPP
+++ b/ECSET.CPP
@@ -457,10 +457,8 @@ void canvas::calculate_object_positions() {
             continue;
         }
         if (is_minimap) {
-            obj->minimap_canvas_x =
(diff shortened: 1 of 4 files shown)


GauChoob:
You should probably split the last 3 commits into separate PR before reviewing

PR #382, opened 2026-02-27
Grassplit

Ben3eeE:


grass.cpp
```diff
@@ -5,29 +5,14 @@
...
+    if (elements.size() >= MAX_GRASS_PICS) {
         external_error("Too many grass pictures in lgr file!");
     }
-    pics[length] = pic;
-    is_up[length] = up;
 
-    length++;
-}
-
-grass::~grass() {
```
Ben3eeE: Why do we just delete the destructor? We still need to delete the pic8?

grass.cpp
```diff
@@ -5,29 +5,14 @@
...
 #include <cmath>
 
-grass::grass() {
-    length = 0;
-    for (int i = 0; i < MAX_GRASS_PICS; i++) {
-        pics[i] = nullptr;
-        is_up[i] = false;
-    }
-}
+grass::grass() { elements.reserve(0); }
```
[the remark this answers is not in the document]
GauChoob: I did it because clang tidy docs say you're supposed to reserve before adding to s vector via loop (i didn't actually run tidy though)

grass.cpp
GauChoob: I thought unique_ptr was supposed to take care of that

PR #399, opened 2026-03-02
Add physics unit tests

Ben3eeE:


GauChoob (commented):
Cool stuff. Haven't reviewed properly yet but just noticed these two things

PR #401, opened 2026-03-03
Show total time in top right corner of menu

Ben3eeE:
Assumes "Player 1" is what you want to see in multiplayer, which makes sense to my brain.
Doesn't work in 640x480 resolution.


GauChoob (changes requested):
Two comments to possibly improve code

PR #412, opened 2026-03-06
Grassstuff3

jgouly:


jgouly:
I tried a few levels with 0.25 zoom but couldn't see the issue the first commit tries to fix. Second commit neeeds justification or constexprs for the 120/50 magic numbers.

PR #425, opened 2026-03-13
fix(ecset): eliminate buglines drawing

Amper:


GauChoob:
I can't test this since I'm travelling but this fix seems a bit troublesome:

At the moment, lines are truncated using (int) + 1 and (int) to round the end of each line segment up/down respectively, to shorten the lines so that two connecting line segments don't produce two chunks for the same vertical row. There is a subsequent check for duplicated chunks but this check has its own line bug issues as well

This fix extends the length of the positive end of the line segment *past* its actual end to fix line bugs resulting from a rounding error between two line segments

Issues:

1) First of all, it's not explained in the PR that we are trying to fix one source of line bugs and not *all* the line bugs, since this isn't the only place in the code where they are produced (I think from reading the code, but not actually tested - the chunk merging step probably also produces line bugs)
2) Next, if two line segments are shaped like a stalagmite or stalagtite, I'm worried about new line bugs being created by this fix with certain vertex values and zooms
3) This is identical to (2), but conceptually this fix graphically increases the sharp point of vertices by 1 pixel in size, which is probably not desired

If we want to fix line bugs, imo we should address the underlying causes instead of patching it over with an unintuitive fix
1) First, we should change the (int) + 1 to use std::ceil to prevent line bugs caused by a vertex with a perfectly integer y value
2) Secondly, if line bugs persist, we could consider storing the original value of the second vertex in a line segment for the purposes of calculating the for y loop:
```
Instead of using std::ceil(r.y) and std::floor(r.y + v.y)
Consider
for(y=std::ceil(r1.y); y<std::floor(r2.y), y++)

Where r2 = r + v, without losing floating point precision from intermediate calculations
```
3) Outside the scope of this specific fix, but to handle other line bugs, you would probably need to store the xpos value of the chunks as a floating point number instead of an int to ensure proper sorting in the sort_xpos function

Imo from a combination of these 3 steps I would expect to get rid of line bugs or almost all of them at least

Issue #55, opened 2025-12-13
Screenshot issues


GauChoob:
1) spam screenshot to drop fps is a bug introduced by mila/hib. Do we want to bring it back?
2) Save into a folder instead of main directory
3) Save as bmp or something instead of pcx
4) Maybe support screenshotting entire level as well instead of just the screen (requires pic8 upgrade to support 2^32 instead of 2^16 size)

PR #432, opened 2026-03-18
Save screenshot as bmp instead of pcx

Ben3eeE:
Implements point 2 and 3 of https://github.com/elmadev/elma-classic/issues/55

### DIFF
--- a/EDITUJ.CPP
+++ b/EDITUJ.CPP
@@ -750,24 +750,7 @@ void editor() {
                 }
 
                 if (c == 'i') {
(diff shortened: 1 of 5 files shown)


GauChoob (changes requested):
This PR is not complete: please also handle screenshot requests originating from the editor (EDITUJ.CPP line 763)

Ben3eeE:
Fixed

GauChoob (approved):
Approved but I don't like the filesystem stuff being in platform. Best imo would be to break it into two functions

PR #446, opened 2026-03-27
Add a replay cache to optimize Merge with and Level replays

Ben3eeE:


include/menu/replay.h
```diff
@@ -0,0 +1,9 @@
+#ifndef REPLAY_MENU_H
+#define REPLAY_MENU_H
+
+#include <string>
```
GauChoob: Technically this #include line is in the wrong commit but also who cares

src/menu/rec_list.cpp
```diff
@@ -12,3 +13,14 @@ std::vector<std::string> rec_list::get_replays() {
     find_close();
     return filenames;
 }
+
+std::vector<std::string> rec_list::replays_for_level(int level_id) {
+    std::vector<std::string> result;
+    for (const auto& filename : get_replays()) {
+        auto header = recorder::read_header(filename);
```
GauChoob: Maybe would be nice to have explicit type for header for readability purposes

src/menu/replay_cache.cpp
```diff
@@ -0,0 +1,144 @@
...
+void replay_cache::wait() {
+    if (worker_.joinable()) {
+        worker_.join();
+    }
+}
+
+bool replay_cache::is_ready() const { return ready_.load(std::memory_order_acquire); }
+
+std::vector<std::string> replay_cache::filenames_for_level(int level_id) const {
+    std::scoped_lock lock(mutex_);
```
GauChoob: imo internal_error or something if !ready_
?

include/menu/replay_cache.h
```diff
@@ -0,0 +1,43 @@
...
+
+    void apply_one(const std::string& filename, int level_id);
+    void apply_pending();
+
+  public:
+    ~replay_cache();
+
+    // Start background scan of rec/ directory
+    void start();
+    // Block until scan completes
```
GauChoob: until initial scan

src/menu/replay_cache.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+replay_cache::~replay_cache() {
+    stop_requested_.store(true, std::memory_order_relaxed);
+    if (worker_.joinable()) {
+        worker_.join();
+    }
+}
+
+void replay_cache::start() {
+    if (worker_.joinable()) {
```
GauChoob: We can probably hiba if ready_ is true or worker is joinable

GauChoob (changes requested):
Looks good in principle. A few comments here and there.

Main one is merging the two different rec/ iterator implementations

Also overall replay_cache is very heavy due to the thread management. Imo it would benefit from at least 30 more lines of comments in the source so that it is faster to understand everything that's going on.

GauChoob (approved):

PR #454, opened 2026-03-30
Translations

2bf3:
- camera global translations
- camera animation state translations
- level time (eddig) translation


jgouly:
Thanks but closing since we translated KIRAJ320 fully now.

PR #349, opened 2026-02-20
Resizeable Window

GauChoob:
Make the window resizable (works everywhere except in EDITTOLT which might be addressed when EDITTOLT is translated and cleaned up)

Resolves https://github.com/elmadev/elma-classic/issues/435
Resolves https://github.com/elmadev/elma-classic/issues/410
Resolves https://github.com/elmadev/elma-classic/issues/202

PR #459, opened 2026-03-30
Resizeable Window Part 1 (Refactoring platform_sdl)

GauChoob:
platform_sdl refactoring prep work

Basically I'm trying to improve the control flow since now we need to update the window via multiple pathways, and there's some duplicate code/logic

### DIFF
--- a/platform_sdl.cpp
+++ b/platform_sdl.cpp
@@ -69,6 +69,19 @@ static void apply_current_palette() {
     }
 }
 
+static void destroy_window() {
+    gl_cleanup();
+
+    if (SDLSurfaceMain) {
+        SDL_DestroyWindowSurface(SDLWindow);
+        SDLSurfaceMain = nullptr;
+    }
+    if (SDLWindow) {
+        SDL_DestroyWindow(SDLWindow);
+        SDLWindow = nullptr;
+    }
+}
+
 static void create_window(int window_pos_x, int window_pos_y, int width, int height) {
     if (EolSettings->renderer() == RendererType::OpenGL) {
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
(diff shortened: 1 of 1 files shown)


jgouly:
Don't really understand the first commit, it just moves stuff but I can't see why. I looked at the full PR too and it's not like we call create_window() in any different places?

jgouly:
For the 2nd commit why can't we reuse platform_resize_window()? Also are you using the gl_renderer or software? crashes for me wil gl_renderer,  not sure why yet.

GauChoob:
First commit isn't directly impacted by this PR, but it was part of the overall window SDL refactoring. This is the second half of simplifying window creation that was already merged in: https://github.com/elmadev/elma-classic/commit/6bc189993533550768d667c0d60882372dc9423d

Overall the 5 platform commits (already merged and in this PR) don't directly impact the resizeable window, but are just simplifying/improving the flow to make resizeable window and future PRs much easier to achieve. I extract the atomic functions instead of having large functions that do multiple things, because this part of the code is kind of heavy and not intuitive. The issue I was mainly running into is when you create a new platform_window function, you have to copy/paste subparts of other functions and make sure that all the parts are in a very specific order. By simplifying like this, the control flow is shorter and simpler.

GauChoob:
Second commit:

We can't use - platform_resize_window runs code that changes the dimensions of the window. Whereas if the user maximizes/resizes the window himself, the dimensions of the window are already changed so we don't want to call the entire platform_resize_window - we just want to recover our window surface without resizing.

window_resized is called directly here https://github.com/elmadev/elma-classic/pull/349/changes/395563f2147ddd81c66de5846bde6e0bbe0b87d3

GauChoob:
> or the 2nd commit why can't we reuse platform_resize_window()? Also are you using the gl_renderer or software? crashes for me wil gl_renderer, not sure why yet.

Works for me, dunno if it's Mac thing. Which commit causes the crash?

jgouly (approved):

PR #462, opened 2026-03-31
One wheel

Ben3eeE:


Ben3eeE:
Rebased to fix conflicts.

GauChoob:
Imo even if we don't reproduce current eol behaviour it would be better to use the current call to get anchor points instead of doing it twice per frame (see my example in my previous comment). The added side effect is wheel actuallly touches polygon before killing bike

I also don't agree with implementation in replay files because it's not accurate

Ben3eeE:
I'll just close this. I was never interested in adding cripples in the first place and my approach is just meeting resistance or being completely ignored. Better to let someone else do it since I have no interest in continuing and adjusting based on the current comments.

PR #473, opened 2026-04-14
Edittolt Part 5

GauChoob:
Fixes related to non-standard resolution

Prep work for eventual resizeable window in the editor

### DIFF
--- a/EDITTOLT.CPP
+++ b/EDITTOLT.CPP
@@ -478,8 +478,11 @@ static void create_lgr_palette_map(unsigned char* pal) {
 }
 
 static void preview_picture(pic8* dest, picture* pict, int x, int y) {
-    if (x < 0 || y < 0 || x >= dest->get_width() || y >= dest->get_height()) {
+    if (x < 0 || y < 0) {
         internal_error("preview_picture out of bounds");
+    }
+    if (x >= dest->get_width() || y >= dest->get_height()) {
+        return;
     }
 
     // Trim width and height to be in-bounds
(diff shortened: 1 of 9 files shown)


Ben3eeE:
Theres still quite many clang tidy warnings including new ones introduced in earlier PRs.

```
../src/editor/window.cpp:241:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:297:9: warning: the 'empty' method should be used to check for emptiness instead of 'length' [readability-container-size-empty]
../src/editor/window.cpp:317:9: warning: the 'empty' method should be used to check for emptiness instead of 'size' [readability-container-size-empty]
../src/editor/window.cpp:322:9: warning: the 'empty' method should be used to check for emptiness instead of 'length' [readability-container-size-empty]
../src/editor/window.cpp:424:17: warning: boolean expression can be simplified by DeMorgan's theorem [readability-simplify-boolean-expr]
../src/editor/window.cpp:661:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:812:5: warning: multiple declarations in a single statement reduces readability [readability-isolate-declaration]
../src/editor/window.cpp:837:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:950:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:1087:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:1215:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:1328:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:1512:11: warning: do not use 'else' after 'return' [readability-else-after-return]
../src/editor/window.cpp:1639:11: warning: do not use 'else' after 'return' [readability-else-after-return]

```

GauChoob:
I can fix them after this PR I guess - I don't think any of these warnings are caused by the commits contained in this PR

jgouly (approved):

PR #474, opened 2026-04-14
fix(menu_play): keep cursor on current selection in level menu

GauChoob:
Fixes a regression


GauChoob:
To reproduce, select any submenu (Replay, Best Times) and then press Esc

jgouly (approved):

PR #476, opened 2026-04-14
Initial EOL public functionality

jgouly:
This does not provide actual online functionality, it is just the public side of the private online code.

Still some tidy up to do, but wanted to open the PR.

TODO:
fixup login event
level length hardcoding

### DIFF
--- a/include/eol/eol.h
+++ b/include/eol/eol.h
@@ -1,6 +1,7 @@
 #ifndef EOL_H
 #define EOL_H
 
+#include "eol/eol_events.h"
 #include "eol/null_protocol.h"
 
 using protocol = null_protocol;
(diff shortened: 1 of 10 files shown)


include/eol/eol.h
```diff
@@ -0,0 +1,21 @@
...
+    eol();
+
+    void connect() { proto.connect(); }
+    void tick() { proto.tick(); }
+
+  private:
+    protocol proto;
+};
+
+extern eol* eol_client;
```
GauChoob: Should be EolClient in this commit, not the next one

include/eol/eol_events.h
```diff
@@ -10,4 +10,13 @@ struct set_kuski_id {
     unsigned int id2;
 };
 
+struct kuski {
```
GauChoob: Hard to differentiate this from a process struct. Needs comment or needs to be moved?

GauChoob (approved):

PR #480, opened 2026-04-17
EOL enter/exit

jgouly:


jgouly:
exit_level() takes a lot (with more to come) parameters. Maybe it should take a `struct struct exit_level` as a parameter, or there's a `struct level_run(?)` that is passed to both `{enter/exit}_level()` to make the API symmetric.

LEJATSZO.CPP
```diff
@@ -451,6 +451,8 @@ long lejatszo(const char* filenev, CameraMode cameramode) {
     double eddig = 0.0;
     reset_event_buffer();
 
+    EolClient->enter_level(filenev, Ptop);
```
jgouly: if Single here too

GauChoob:
Imo it might make sense to eventually have a struct that just contains all the local play variables (time, win/lose, normal / apple battle, cripples, etc) that we can use in the function and also pass to the eol function

(Would also be a change needed for eventual wasm if we do that)

GauChoob (approved):
Imo can merge once you're satisfied with it - we can always adjust the endpoint as we go/need

PR #489, opened 2026-04-21
Move platform code to src folder and random stuff

Ben3eeE:


include/menu/dialog.h
```diff
@@ -1,7 +1,7 @@
 #ifndef MENU_DIALOG_H
 #define MENU_DIALOG_H
 
-#include <platform_impl.h>
+#include "platform/implementation.h"
```
GauChoob, line 4: I guess header #define should be PLATFORM_IMPLEMENTATION_H or IMPLEMENTATION_H (or other if we rename it)

GauChoob (approved):
Approved with some comments to address:

I find implementation.h a little bit vague as a term, but I don't really know what to call it

PR #492, opened 2026-04-24
Log

Ben3eeE:


include/log.h
```diff
@@ -0,0 +1,50 @@
...
+    void write(LogLevel lvl, std::source_location loc, std::string msg);
+
+    logger(const logger&) = delete;
+    logger& operator=(const logger&) = delete;
+
+  private:
+    logger();
+    ~logger();
+
+    std::FILE* file_ = nullptr;
```
jgouly: We've been only using field_ when we want a method with the same name

src/log.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    auto now = std::chrono::floor<std::chrono::seconds>(std::chrono::system_clock::now());
+    std::fputs(std::format("[{:%H:%M:%S}] [{}] {}:{}: {}\n", now, level_name(lvl),
+                           basename(loc.file_name()), loc.line(), msg)
+                   .c_str(),
+               file_);
+    std::fflush(file_);
+
+    if (Console) {
+        Console->add_line(std::format("[{}] {}", level_name(lvl), msg), console::LineType::System);
```
jgouly: I thought this was going to print the name of the level you were playing :D

Ben3eeE:
No idea if this works on Windows but I made the compiler happy.
Does not distinguish between `menu/window.cpp` and `editor/window.cpp` because we only log the basename of the file.

src/log.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    auto now = std::chrono::floor<std::chrono::seconds>(std::chrono::system_clock::now());
+    std::fputs(std::format("[{:%H:%M:%S}] [{}] {}:{}: {}\n", now, log_level_name(lvl),
+                           basename(loc.file_name()), loc.line(), msg)
+                   .c_str(),
+               file);
+    std::fflush(file);
+
+    if (Console) {
+        Console->add_line(std::format("[{}] {}", log_level_name(lvl), msg),
```
GauChoob: Can console.h default to Mode::Console instead of Mode::Chat if DEBUG is enabled? I was confused why nothing was showing

src/log.cpp
```diff
@@ -0,0 +1,59 @@
+#include "log.h"
+
+#include "eol/console.h"
+#include <chrono>
+#include <cstring>
+
+static const char* log_level_name(LogLevel lvl) {
```
GauChoob: I kind of find log_level_name confusing still. "level_name" calls out to me :P

How about to_string(LogLevel lvl)?

src/log.cpp
```diff
@@ -0,0 +1,59 @@
...
+    auto pos = path.find_last_of("/\\");
+    return pos == std::string_view::npos ? path : path.substr(pos + 1);
+}
+
+logger& logger::instance() {
+    static logger inst;
+    return inst;
+}
+
+logger::logger() { file = std::fopen("eol.log", "w"); }
```
GauChoob: Is it intentional that we rewrite the log file vs appending? Which approach do we want?

src/log.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (!file) {
+        return;
+    }
+
+    auto now = std::chrono::floor<std::chrono::seconds>(std::chrono::system_clock::now());
+    std::fputs(std::format("[{:%H:%M:%S}] [{}] {}:{}: {}\n", now, log_level_name(lvl),
+                           basename(loc.file_name()), loc.line(), msg)
+                   .c_str(),
+               file);
+    std::fflush(file);
```
GauChoob: Is fflush associated with a performance cost? Should we not flush, or only flush here if DEBUG?

include/log.h
```diff
@@ -0,0 +1,50 @@
+#ifndef LOG_H
```
GauChoob, line 1: I do think we should have a commit to replace the std::cout in profiler.cpp because otherwise we have two competing logging systems

include/log.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <cstdio>
+#include <format>
+#include <mutex>
+#include <source_location>
+#include <string>
+
+enum class LogLevel { Debug, Info, Warn, Error, Fatal };
+
+class logger {
```
GauChoob, line 36: Maybe a quick comment that logger is a singleton and to access via instance()? It's obvious but at the same time you could miss it. Dunno

src/log.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+logger::logger() { file = std::fopen("eol.log", "w"); }
+
+logger::~logger() {
+    if (file) {
+        std::fclose(file);
+    }
+}
+
+void logger::write(LogLevel lvl, std::source_location loc, std::string msg) {
```
GauChoob: It's intentional that we don't write to std::out in DEBUG mode?

GauChoob (commented):
Looks really good. A few tiny nitpicks and a few conceptual questions that I asked

src/log.cpp
jgouly: I think appending could be useful so you don't just resttart and lose stuff?

src/log.cpp
Ben3eeE: Yes, you can tail the logfile if you want output in a terminal.

src/log.cpp
Ben3eeE: I don't want to build log rotation to avoid users getting extremely large log files because we append.

include/log.h
Ben3eeE: No one except the one exception of internal_error will ever need to know about this detail. Everyone contributing will use the available macro to log something.

include/log.h
Ben3eeE: Like I said in Discord:

> Yes it could. I didn't touch it because this logs to in game console and file only, no logging to stdout and I didn't know if that was something that was wanted to preserve

I don't know what is desired.

src/log.cpp
Ben3eeE: It can but that is not how the console was designed when this mode was added.

src/log.cpp
Ben3eeE: If we don't flush the log may be incomplete and not give us understanding of what happened.

src/log.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+logger& logger::instance() {
+    static logger inst;
+    return inst;
+}
+
+logger::logger() {
+    namespace fs = std::filesystem;
+
+    constexpr const char* LOG_FILENAME = "eol.log";
```
GauChoob: Is `constexpr const char LOG_FILENAME[] = "eol.log";` better?

src/log.cpp
```diff
@@ -0,0 +1,76 @@
...
+    constexpr unsigned int MAX_LOG_FILE_SIZE = 5 * 1024 * 1024; // 5 MiB
```
GauChoob: 5 MB * 6 = 30 MB of logs seems a bit excessive... maybe 3*1 MB or something like that?

Dunno, curious what joey thinks

src/log.cpp
```diff
@@ -1,25 +1,10 @@
 #include "log.h"
 
+#include "eol/console.h"
 #include <chrono>
 #include <cstring>
 #include <filesystem>
 
-static const char* log_level_name(LogLevel lvl) {
```
GauChoob: I think the deletion of log_level_name was supposed to be a fixup of the previous commit

src/eol/console.cpp
```diff
@@ -71,6 +71,17 @@ void console::register_console_commands() {
     register_command("clear", [this](std::string_view) { clear(); });
     register_command("dev", [this](std::string_view) { mode = Mode::Console; });
     register_command("chat", [this](std::string_view) { mode = Mode::Chat; });
+    register_command("log", [this](std::string_view text) {
+        if (text.empty()) {
+            show_log_lines = !show_log_lines;
+        } else if (auto val = parse_bool(text)) {
+            show_log_lines = *val;
+        } else {
+            add_line(std::format("invalid value: {}", text), LineType::System);
```
GauChoob, line 80: This isn't directly related to this PR, but it doesn't make sense for me for invalid commands such as !log asdasd to silently fail if we aren't in dev mode, since LineType::System only shows up if !dev.

Maybe we can make an Issue to split LineType::System into LineType::Dev and LineType::Warning or something like that? The intent of System is unclear imo

src/log.cpp
```diff
@@ -19,14 +19,14 @@ logger::logger() {
     namespace fs = std::filesystem;
 
     constexpr const char* LOG_FILENAME = "eol.log";
-    constexpr unsigned int MAX_LOG_FILE_SIZE = 5 * 1024 * 1024; // 5 MiB
-    constexpr unsigned int MAX_KEPT_LOG_FILES = 5;
+    constexpr std::size_t MAX_LOG_FILE_SIZE = 5 * 1024 * 1024; // 5 MiB
```
GauChoob: We use "size_t" instead of "std::size_t" everywhere else. If we are switching to std::size_t then we should make an Issue to convert it everywhere else

GauChoob (approved):
Looks pretty much ready. A few teeny comments but after those are addressed I think can merge. Two some details that aren't directly related to this PR that can be converted into issues if they aren't addressed here

src/log.cpp
jgouly: We're basically never going to reach this, and 30MB is nothing now a days :P

PR #493, opened 2026-04-25
Eol settings runtime layer

Ben3eeE:


GauChoob (approved):
PR looks fine, although I do find the word "persist_" a bit heavy over set. If there was a less heavy word for set vs persist (e.g. save_default_lgr_name() and set_default_lgr_name()) I would prefer it

jgouly:
save/set seem too close. Could have set_saved_foo()? But we only using this in the menu_options file, everywhere else will use set_foo() (probably), and it's clear what each one does.

PR #494, opened 2026-04-25
Move editor code and gl_renderer

Ben3eeE:


jgouly (approved):
If we were so inclined we'd somehow abstract the gl_renderer from SDL.. maybe in 5y.

PR #496, opened 2026-04-27
Resizeable Window Part 3/4

GauChoob:


src/platform/sdl/gl_renderer.cpp
```diff
@@ -185,7 +185,7 @@ int gl_init(SDL_Window* sdl_window, int width, int height, int pitch) {
 
     setup_render_state();
 
-    return 0;
+    return;
```
Ben3eeE: redundant return from void function

src/platform/sdl/gl_renderer.cpp
```diff
@@ -232,7 +232,19 @@ int gl_resize(int width, int height, int pitch) {
...
+
+void gl_setup(SDL_Window* sdl_window, int width, int height, int pitch) {
+    if (!GLContext) {
+        gl_init(sdl_window, width, height, pitch);
+        return;
+    }
+
+    if (FrameWidth != width || FrameHeight != height) {
+        gl_resize(width, height, pitch);
+        return;
```
Ben3eeE: redundant return from void function

GauChoob: Fixed a total of 3 redundant returns in gl_renderer.cpp

Ben3eeE:
I don't know why we are merging init and resize into a single function. It's a strange API to call a function called setup to resize. Because it's not doing any setup at all.

I see the benefit you are going for but there has to be a better solution because this is just confusing API design. Also is this really related to resize? There was a PR that was kinda approved already that worked except for one missing patch. Why add on top of it now instead of doing this in a follow up PR? Is this really mandatory for the resize work to get merged?

Ben3eeE:
Thanks for the explanation, I get what you're going for now. I still don't think we should collapse `gl_init` and `gl_resize` into one function though, and I think we can get the simplification you want without doing that.

What bothers me about `gl_setup` isn't really the merge, it's that it decides what to do based on whether `GLContext` is null. But `setup_renderer` already knows if we're in `create_window` or `window_resized`, those are different code paths. So we're throwing away information at the call site and re-deriving it from a global inside the GL module. And the name "setup" reads like init to me, so seeing it called for a resize is a bit confusing every time.

I'd keep `gl_init` and `gl_resize` separate, and just dedupe the thing that's actually duplicated, which is the software path `SDL_GetWindowSurface` block. I pushed it as [`window_p3_alt`](https://github.com/elmadev/elma-classic/tree/window_p3_alt) if you want to try it. The end state for `setup_surfaces` in the next commit is the same, it just calls `resize_renderer` instead.

`include/platform/sdl/gl_renderer.h`:

```diff
-int gl_init(SDL_Window* sdl_window, int width, int height, int pitch);
+void gl_init(SDL_Window* sdl_window, int width, int height, int pitch);
+void gl_resize(int width, int height, int pitch);
 void gl_upload_frame(const unsigned char* indices, int pitch);
 void gl_update_palette(const void* palette);
 void gl_present();
-int gl_resize(int width, int height, int pitch);
 void gl_cleanup();
```

`src/platform/sdl/gl_renderer.cpp`: drop the unused `int` returns from both. They always returned 0 anyway and the checks at the call site were dead.

`src/platform/sdl/sdl.cpp`:

```diff
-static void initialize_renderer() {
+static void refresh_software_surface() {
+    SDLSurfaceMain = SDL_GetWindowSurface(SDLWindow);
+    if (!SDLSurfaceMain) {
+        internal_error(SDL_GetError());
+    }
+}
+
+static void initialize_renderer(int width, int height) {
     if (EolSettings->renderer() == RendererType::OpenGL) {
-        if (gl_init(SDLWindow, SCREEN_WIDTH, SCREEN_HEIGHT, SDLSurfacePaletted->pitch) != 0) {
-            internal_error("Failed to initialize OpenGL renderer");
-        }
-
-        SDLSurfaceMain = nullptr; // Not used in GL mode
+        gl_init(SDLWindow, width, height, SDLSurfacePaletted->pitch);
+        SDLSurfaceMain = nullptr; // Not used in GL mode
     } else {
-        SDLSurfaceMain = SDL_GetWindowSurface(SDLWindow);
-        if (!SDLSurfaceMain) {
-            internal_error(SDL_GetError());
-        }
+        refresh_software_surface();
     }
 }

 static void resize_renderer(int width, int height) {
     if (EolSettings->renderer() == RendererType::OpenGL) {
-        if (gl_resize(width, height, SDLSurfacePaletted->pitch) != 0) {
-            internal_error("Failed to resize OpenGL renderer");
-        }
-        SDLSurfaceMain = nullptr; // Not used in GL mode
+        gl_resize(width, height, SDLSurfacePaletted->pitch);
+        SDLSurfaceMain = nullptr; // Not used in GL mode
     } else {
-        SDLSurfaceMain = SDL_GetWindowSurface(SDLWindow);
-        if (!SDLSurfaceMain) {
-            internal_error(SDL_GetError());
-        }
+        refresh_software_surface();
     }
 }
```

Then `setup_surfaces` in the next commit calls `resize_renderer`, since every site that goes through it is post-creation, and `create_window` keeps calling `initialize_renderer` directly. Same dedup, no global-state dispatch, each GL function still does one thing.

The reason I think this matters: I get that the goal is fewer entry points so things stop breaking on small changes. But I think `if (!GLContext)` style runtime dispatch is actually the kind of thing that causes "slightest change breaks something" over time, not what fixes it. Each caller stops being able to predict what the function does without reading its body and tracing global state. With separate `gl_init` and `gl_resize`, each has a narrow contract you can check at the call site, and a wrong call fails visibly instead of silently doing something else. Pushing the decision back to where the lifecycle is actually known (sdl.cpp) removes a class of footgun without giving up the call-site simplicity, `setup_surfaces` looks the same to callers either way.

Also, this refactor isn't really tied to the resize feature, can it just be its own PR after the previously approved one lands? Would unblock the resize work.

Ben3eeE (approved):

PR #501, opened 2026-04-28
Edittool Part 2

GauChoob:


level.cpp
```diff
@@ -340,6 +340,53 @@ sprite* level::get_closest_sprite(double x, double y, double* distance) {
...
+void level::get_closest_entity(int pixel_x, int pixel_y, polygon** poly, int* vertex_index,
+                               object** obj, sprite** spr) {
+    double x = pixel_to_meter_x(pixel_x);
+    double y = pixel_to_meter_y(pixel_y);
+
+    double distance_vertex;
+    double distance_object;
+    double distance_sprite;
+    if (poly && vertex_index) {
+        if (vertex_index == nullptr) {
```
jgouly: Im assuming line 351 shouldn't have the ` && vertex_index`, I can remove that if you agree

GauChoob: Yeah my mistake, thanks

jgouly (approved):

PR #517, opened 2026-05-04
Basic spy

jgouly:


src/renderer/render.cpp
```diff
@@ -343,6 +343,20 @@ static void render_minimap(bool player1, pic8* pic, double camera_turn_phase, ve
...
     }
 
+    if (EolSettings->show_others()) {
+        for (const kuski& ku : EolClient->kuskis()) {
+            const spy_data* k = ku.spy_data();
+            if (!k) {
+                continue;
+            }
+
+            vect2 other_pos = k->mot.bike.r - bottomleft_corner;
```
GauChoob: Imo Inwould prefer k_pos and k_x here (or better, kus_pos and kus_x) since "other" is kind of used to refer to the second player in Multiplayer mode

include/eol/eol_types.h
```diff
@@ -1,12 +1,31 @@
 #ifndef EOL_TYPES_H
 #define EOL_TYPES_H
 
+#include "LEJATSZO.H"
 #include "fs_utils.h"
+#include "physics_init.h"
+#include <cstdint>
+#include <optional>
+
+class level;
```
GauChoob: I'm not at computer but I'm surprised we need to declare level here. This is used?

GauChoob (approved):

include/eol/eol.h
```diff
@@ -29,13 +29,15 @@ class eol {
     void set_table(TableType);
     void render_table(pic8& dest, abc8& title_font, abc8& data_font) const;
 
+    const std::vector<kuski>& kuskis() { return kuskis_; }
```
Ben3eeE, line 32: These are only online kuskis. Battle needs all. How will that be named?

jgouly: I see two options:

- add an 'online' / 'is_player' field to kuski and filter them out where needed
- second vector and add some helpers like `get_kuski_name()` that will iterate both vecs

I don't think battle or other code or anything outside of `eol`class/folder will specifically need to iterate over ALL kuskis.

Issue #226, opened 2026-01-26
Delete affine_pic? and replace with pic8?


GauChoob:
It would also mean we can get rid of the bike chopping (lgrfile::chop_bike) and have faster/better rendering

PR #518, opened 2026-05-04
refactor(affine_pic): delete affine_pic class and replace it with pic8

GauChoob:
We could simplify the code by deleting the class affine_pic, which basically only serves to rapidly calculate gpixel() using fixed-point integers. Since we've replaced the fixed-point integers with doubles, there isn't much benefit to keeping the affine_pic class around. We can simply use pic8. The only visible change should be that lgrs not longer crash if a bike asset is larger than 256 in width/height.

I had to add a bit of const here and there to the pic8 class since joey added const to some of the function params (I guess for online bikes? not really sure why he did that)

Resolves the first half of https://github.com/elmadev/elma-classic/issues/226, but imo we don't necessarily need to do the second half because it might have a negative performance impact


jgouly:
Could be worth a quick profile just to be sure

GauChoob:
Did a diff and identical picture is rendered

GauChoob:
Profile test:
Rendering one entire bike takes about:
About 0.143 milliseconds in this branch
About 0.107 milliseconds in main branch

GauChoob:
In Release:
0.015 milliseconds in this branch
0.023 milliseconds in main branch

So it seems to even have a speedup when optimized

jgouly (approved):

jgouly:
and yes the const I added was for online bikes, since we iterate over them as 'const', so had to rework the signatures to allow passing const!

PR #520, opened 2026-05-05
fix(ci): widen vs-version range to include 18.x

Ben3eeE:
windows-latest is vs-version 18 these days apparently.

https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/#windows-server-2025-with-visual-studio-2026-image-now-available-for-github-hosted-runners


GauChoob:
Why not just 18-19 instead of 17-19

jgouly (approved):

PR #522, opened 2026-05-07
Zoom front grass

GauChoob:
The long awaited last missing feature in zoom

(skipped if Pics in background is enabled)

### DIFF
--- a/canvas.cpp
+++ b/canvas.cpp
@@ -1589,17 +1589,17 @@ void canvas::create_canvases() {
 }
 
 node_finder::node_finder(canvas* src) {
-    current_x = -1;
+    current_x_ = -1;
     current_y = -1;
     current_node = nullptr;
     source = src;
 }
 
 canvas_chunk_node* node_finder::get_chunk(int x, int y) {
-    if (y == current_y && x >= current_x) {
+    if (y == current_y && x >= current_x_) {
         // Shortcut to iterate to the right of our last node
-        while (x >= current_x + current_node->width) {
-            current_x += current_node->width;
+        while (x >= current_x_ + current_node->width) {
+            current_x_ += current_node->width;
             current_node = current_node->next;
         }
     } else {
@@ -1609,15 +1609,15 @@ canvas_chunk_node* node_finder::get_chunk(int x, int y) {
         }
         current_y = y;
         current_node = source->rows_linked[y];
-        current_x = source->rows_x1[y];
+        current_x_ = source->rows_x1[y];
 
         // Make sure canvas is unused so far
-        if (current_x > 10) {
-            internal_error("node_finder current_x > 10");
+        if (current_x_ > 10) {
+            internal_error("node_finder current_x_ > 10");
         }
 
-        while (x >= current_x + current_node->width) {
-            current_x += current_node->width;
+        while (x >= current_x_ + current_node->width) {
+            current_x_ += current_node->width;
             current_node = current_node->next;
         }
     }
--- a/canvas.cpp
+++ b/canvas.cpp
@@ -1446,15 +1446,39 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
     return count;
 }
 
-void canvas::create_front_grass() {
-    node_finder* finder = new node_finder(this);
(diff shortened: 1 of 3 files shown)


canvas.cpp
```diff
@@ -1446,15 +1446,39 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
...
+            cur_node = cur_node->next;
+        }
+    }
+}
 
+void canvas::create_front_grass() {
     // A crutch that we use to avoid keeping proper track of our linked lists,
     // by making sure our chunks don't get merged together.
-    bool alternate_distance = false;
+    int alternate_distance = false;
```
Ben3eeE, line 1470: int = false

canvas.cpp
```diff
@@ -1482,18 +1506,17 @@ void canvas::create_front_grass() {
...
                             // Move qgrass to front
                             alternate_distance = !alternate_distance;
-                            if (alternate_distance) {
-                                draw_pixels(cur_node->pixels, 223, x1, x1 + count - 1, i,
-                                            Clipping::Unclipped);
-                            } else {
-                                draw_pixels(cur_node->pixels, 224, x1, x1 + count - 1, i,
-                                            Clipping::Unclipped);
+                            const int distance = alternate_distance ? 223 : 224;
+                            const int x2 = x1 + count - 1;
```
Ben3eeE, line 1516: shadowed x2 is confusing.

canvas.h
```diff
@@ -20,13 +20,14 @@ struct segment;
...
     canvas* source;
-    int current_x;
+    int current_x_;
     int current_y;
     canvas_chunk_node* current_node;
 
   public:
     node_finder(canvas* src);
     canvas_chunk_node* get_chunk(int x, int y);
+    int current_x() const { return current_x_; };
```
Ben3eeE, line 30: extra semicolon

canvas.cpp
```diff
@@ -1446,15 +1446,39 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
...
+void canvas::thicken_front_grass(int thickness, int x1, int x2, int y) {
+    node_finder finder(this);
+
+    for (int i = y; i > y - thickness; i--) {
+        canvas_chunk_node* cur_node = finder.get_chunk(x1, i);
+        int j = finder.current_x();
+        while (j <= x2) {
+            if (cur_node->pixels == *QgrassTextureId && cur_node->distance > 500) {
+                int j1 = std::max(x1, j);
+                int j2 = std::min(x2, j + cur_node->width);
```
Ben3eeE, line 1458: draw_pixels x_right is inclusive? Every other call site is width - 1.

Why is it not
```c++
int j2 = std::min(x2, j + cur_node->width - 1);
```

canvas.cpp
```diff
@@ -1446,15 +1446,39 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
...
+                canvas::draw_pixels(*QgrassTextureId, 200, j1, j2, i, Clipping::Unclipped);
+            }
+            j += cur_node->width;
```
Ben3eeE, line 1461: Doesn't draw_pixels mutate the chunk list (split cur_node or unlink it via merge)?

create_front_grass snapshots width and next before drawing for that reason. Why don't we do the same here?

canvas.cpp
```diff
@@ -1446,15 +1446,39 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
     return count;
 }
 
-void canvas::create_front_grass() {
-    node_finder* finder = new node_finder(this);
+void canvas::thicken_front_grass(int thickness, int x1, int x2, int y) {
+    node_finder finder(this);
```
Ben3eeE, line 1450: Why a fresh node_finder per call? consecutive_front/back pass it as pointer.

canvas.cpp
```diff
@@ -1446,15 +1446,39 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
...
+
+    for (int i = y; i > y - thickness; i--) {
```
Ben3eeE, line 1452: Does the i == y iteration ever do anything? The caller paints row y at distance 223/224 right before, so the > 500 guard would always skip?

PR #524, opened 2026-05-07
Render Battel status (F10)

### LINKED ISSUE #477: Players online table improvements
- Internals show as QWQUUXXX, but in EOL it's "internal XX".
- Other levels need `.lev` appending. They show as "XYZ", but in EOL they show as "XYZ.lev"
- Current player doesn't show up in players online table, eol_table class API needs reworking somehow

Ben3eeE:
Fixes https://github.com/elmadev/elma-classic/issues/477

### DIFF
--- a/include/eol/eol.h
+++ b/include/eol/eol.h
@@ -4,6 +4,7 @@
 #include "eol/eol_events.h"
 #include "eol/eol_table.h"
 #include "eol/protocol.h"
+#include "eol_settings.h"
 #include <optional>
 #include <ranges>
 #include <string_view>
@@ -42,6 +43,10 @@ class eol {
         return kuskis_ |
                std::views::filter([](const kuski& k) { return k.is_online && k.is_player; });
     }
+    void toggle_battle_status() {
+        EolSettings->set_show_battle_status(!EolSettings->show_battle_status());
+    }
+    void render_battle_status(pic8& dest, abc8& font) const;
 
   private:
     void sync_players_online_table();
--- a/src/eol/eol.cpp
+++ b/src/eol/eol.cpp
@@ -1,14 +1,19 @@
 #include "eol/eol.h"
 #include "eol/status_messages.h"
 #include "eol/console.h"
+#include "abc8.h"
 #include "eol_settings.h"
+#include "level.h"
 #include "log.h"
+#include "pic8.h"
 #include "platform/implementation.h"
 #include "platform/utils.h"
 #include <algorithm>
-#include <cstring>
 #include <chrono>
+#include <cstring>
 #include <format>
+#include <string>
+#include <string_view>
 
 static kuski* get_kuski(std::vector<kuski>& kuskis, unsigned int id) {
     for (kuski& k : kuskis) {
@@ -18,6 +23,16 @@ static kuski* get_kuski(std::vector<kuski>& kuskis, unsigned int id) {
     }
 
     return nullptr;
+}
+
+static std::string format_level(std::string_view level) {
+    std::string with_ext = std::format("{}.lev", level);
+    auto idx = get_internal_index(with_ext.c_str());
+
+    if (idx.has_value()) {
+        return std::format("internal {:02}", *idx);
+    }
+    return with_ext;
 }
 
 eol::eol()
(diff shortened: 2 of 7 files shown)


src/eol/eol.cpp
```diff
@@ -7,8 +7,15 @@
 #include "platform/utils.h"
 #include <algorithm>
 #include <cstring>
+#include "abc8.h"
```
jgouly: wrong ordering here

src/eol/eol.cpp
```diff
@@ -220,3 +227,124 @@ const struct spy_data* kuski::spy_data() const { return data ? &*data : nullptr;
...
+        out = std::format("{} in {} by {} starts in {} ({})", type_text, level_text, designer,
+                          time_text, duration_text);
+    } else if (current_battle->type == BattleType::HourTT) {
+        out = std::format("{} by {} ends in {}", type_text, designer, time_text);
+    } else {
+        out = std::format("{} in {} by {} ends in {} / {}", type_text, level_text, designer,
+                          time_text, duration_text);
+    }
+
+    const int y = 15 + font.line_height() * (1 + EolSettings->chat_lines());
```
Ben3eeE, line 344: This is weird but I copied it from eol I promise.

src/eol/eol.cpp
```diff
@@ -220,3 +227,124 @@ const struct spy_data* kuski::spy_data() const { return data ? &*data : nullptr;
...
+    std::string with_ext = std::format("{}.lev", level);
+    auto idx = get_internal_index(with_ext.c_str());
+
+    if (idx.has_value()) {
+        return std::format("internal {:02}", *idx);
+    }
+    return with_ext;
+}
+
+static std::string format_hms(int seconds) {
```
GauChoob, line 239: Should go into util.cpp beside centiseconds_to_string? Should group all the time->string functions together

src/eol/eol.cpp
```diff
@@ -220,3 +227,124 @@ const struct spy_data* kuski::spy_data() const { return data ? &*data : nullptr;
...
+    for (auto [flag, text] : cripples) {
+        if (attrs & flag) {
+            out += text;
+        }
+    }
+
+    out += "battle";
+
+    if (out.front() >= 'a' && out.front() <= 'z') {
+        out.front() -= 'a' - 'A';
```
GauChoob: Semantics but how about `out.front() += 'A' - 'a';`

lol

src/eol/eol.cpp
GauChoob: Maybe time to update equation?

src/eol/eol.cpp
Ben3eeE: I say no because it's only used in one place today. There is no value to create a shared function if there is only one user.

src/eol/eol.cpp
Ben3eeE: Not something I want to do now. Can be done later potentially.

src/eol/eol.cpp
GauChoob: In future we will end up with 3 very slightly different time format functions and then will be unable to reconcile them. So imo I disagree with you because it's not future-proofing (but also not a big deal). imo ask joey ask tie-breaker and then move on

src/eol/eol.cpp
```diff
@@ -220,3 +225,124 @@ const struct spy_data* kuski::spy_data() const { return data ? &*data : nullptr;
 void kuski::add_spy_data(const struct spy_data& sd) { data = sd; }
 
 void kuski::clear_spy_data() { data = std::nullopt; }
+
+static std::string format_level(std::string_view level) {
```
jgouly: Can you place this higher up and add a new commit to use it in sync_players_online_table()? Then #477 can be closed

include/eol/eol.h
```diff
@@ -56,6 +58,7 @@ class eol {
     std::optional<battle> current_battle;
     eol_table* cur_table;
     eol_table players_online_table;
+    bool show_battle_status = true;
```
Ben3eeE: Need to check if this should be in eol settings actually

GauChoob (approved):

src/eol/eol.cpp
jgouly: My tie breaker is just leave as-is, until we have reason to share

PR #525, opened 2026-05-08
feat(eol): add F2/shift-F2 to spy a kuski

jgouly:


jgouly:
This needs to clear when exiting a level.

src/eol/eol.cpp
```diff
@@ -193,6 +178,7 @@ void eol::enter_level(const char* level_name, const level* lev) {
 
 void eol::exit_level(const char* level_name, double time, int apple_count, int level_apple_count,
                      bool dead) {
+    spy_kuski_id = std::nullopt;
```
Ben3eeE: I have used `.clear()` in battle code. So now we have a battle about what to use. I see chat is also assigning `std::nullopt`. Other places in code only use `std::nullopt` as a possible return value.

jgouly: `.reset()`?

Ben3eeE: Yes I meant that 👍 
```c++
current_battle.reset();
```

jgouly: I actually prefer = nullopt, feels more explicit, but I will change it

Ben3eeE: Oh, my preference is the `.reset()` :cow: but we can use `std::nullopt` instead, I don't mind. But we should update all places to be the same thing and align on which we use so we don't mix them around.

We also have a mix of `if (!blabla)` and `if(!blala.has_value())` in the code :cow: I don't have a strong opinion on which we use, my strong opinion is on mixing approaches.

Consistency is important.

Ben3eeE (approved):

src/eol/eol.cpp
```diff
@@ -250,11 +251,48 @@ void eol::render_table(pic8& dest, abc8& title_font, abc8& data_font) const {
...
+            continue;
+        }
+
+        if (found_current) {
+            spy_kuski_id = k.id;
+            StatusMessages->add(std::format("now observing {}", k.nick));
+            return;
+        }
+
+        if (*spy_kuski_id == k.id) {
```
GauChoob: Should this check for found_current be earlier than 
```
        if (!k.spy_data()) {
            continue;
        }
```
?

Sometimes the person you are spying has Esced the level so they might have no spy_data() but you are still focused on them (e.g. if they Esc and then re-enter the level)

src/eol/eol.cpp
GauChoob: I like reset() but I also don't really care

PR #526, opened 2026-05-08
Zoom front grass v2

GauChoob:
I just completely rewrote the thicken_front_grass function so I'm closing the old PR and creating a new one.

For bene:

Even though the previous function seemed to work without issues, draw_pixels() can merge nodes together and it's hard to say definitively that there won't ever be any bugs.

Fundamentally, draw_pixels() changes our linked list and so it's difficult to keep track of our list - that's why create_front_grass() alternates between a distance of 223 and 224 to prevent modifications to the linked list. Not only is this kind of a dumb way of doing it, thicken_front_grass() cannot use this solution since it might be called twice on the same chunk of grass.

Lastly, since the linked list might be changed by draw_pixels(), using the node_finder helper class is also risky. (This doesn't apply to Balazs' usage of node_finder since that one iterates over an immutable row)

So the solution is to call draw_one_chunk() which is the child function of draw_pixels(). This allows us to keep track of modifications to the linked list inline within our function. This also has the added benefit of decreasing the number of times we need to iterate through our linked list by a factor of 2, speeding up the function.

tldr: I rewrote the function, please re-review

### DIFF
--- a/canvas.cpp
+++ b/canvas.cpp
@@ -1446,15 +1446,70 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
     return count;
 }
 
-void canvas::create_front_grass() {
-    node_finder* finder = new node_finder(this);
+void canvas::thicken_front_grass(int thickness, int x1, int x2, int y) {
+    // Skip the top row of front grass (handled in create_front_grass())
+    for (int i = y - 1; i > y - thickness; i--) {
+        // Find the starting chunk
+        canvas_chunk_node* prev_node = nullptr;
+        canvas_chunk_node* cur_node = rows_linked[i];
+        int j = rows_x1[i];
 
+#ifdef DEBUG
+        if (i < 0 || i >= pixel_height) {
+            internal_error("canvas::thicken_front_grass invalid y");
+        }
+        if (j > 10) {
+            internal_error("canvas::thicken_front_grass j > 10");
+        }
+#endif
+
+        while (x1 >= j + cur_node->width) {
+            j += cur_node->width;
+            prev_node = cur_node;
+            cur_node = cur_node->next;
+            if (!cur_node) {
+                internal_error("canvas::thicken_front_grass !cur_node!");
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


changelog.txt
```diff
@@ -54,10 +54,11 @@
 * Screenshots are now saved as bmp instead of pcx
 * error.txt has been replaced with eol.log
 * You can select a default gravity and animation via right click when using the Create Food tool in the editor
+* Front grass in front of the bike is properly zoomed
 
 ## Minor Bug Fixes
 * Enforce lowercase LGR filenames to fix loading issues on case-sensitive filesystems
-* Foreground grass was accidentally inefficiently calculated, causing increased load time
+* Front grass  in front of the bike was accidentally inefficiently calculated, causing increased load time
```
Ben3eeE: nit: double space

canvas.cpp
```diff
@@ -1446,15 +1446,70 @@ static int consecutive_front(int x1, int x2, int y, node_finder* finder) {
...
+                internal_error("canvas::thicken_front_grass !cur_node!");
+            }
+
+            int j_next = j + cur_node->width;
+            canvas_chunk_node* next_node = cur_node->next;
+            if (cur_node->pixels == *QgrassTextureId && cur_node->distance > 500) {
+                bool prev_can_merge = false;
+                int j1 = std::max(x1, j);
+                int j2 = std::min(x2, j_next - 1);
+                prev_node = draw_one_chunk(cur_node, *QgrassTextureId, j, j1, j2, 200, prev_node,
```
Ben3eeE, line 1487: Should there be a comment why this is not `draw_pixels`? This information is currently only in the PR body.

Ben3eeE:
Didn't look in detail yet but I had two quick comments :bear:

Ben3eeE (approved):

canvas.cpp
GauChoob: It could, but since we are iterating by chunk imo draw_one_chunk also kind of makes sense in its own context

PR #529, opened 2026-05-08
Editor search

### LINKED ISSUE #503: Editor open dialog type to search
Needed for feature compatibility with eol.

Can probably be genetic so works on pictures also

Ben3eeE:
Fixes https://github.com/elmadev/elma-classic/issues/503

### DIFF
--- a/include/util/util.h
+++ b/include/util/util.h
@@ -29,6 +29,9 @@ bool is_filename_char(unsigned char c);
 // If `compact` is true, formats as: x:xx or x:xx:xx.
 void centiseconds_to_string(int time, char* text, bool show_hours = false, bool compact = false);
 
+// Length of the common case-insensitive prefix of two C strings.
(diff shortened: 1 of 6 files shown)


Ben3eeE:
I tried sharing code with menu but it was more complex to make shared function. This is the best I could come up with.

GauChoob:
Didn't review fully, but qgrass is intentionally placed at the end of the textures list because the first two textures are used to override the default foreground/background texture when fg == bg, or when the texture name is not found (missing custom lgr file or wrong lgr file). Placing qgrass at the end of the list avoids qgrass from being used as a default texture (except in the rare case where there are only 2 total textures + QGRASS)

So basically this change can lead to unintended graphical changes.

GauChoob:
For the above, see discard_missing_lgr_assets()

GauChoob (approved):

PR #533, opened 2026-05-09
fix(lejatszo): reset `WhoDiedFirst` in single player mode

GauChoob:


LEJATSZO.CPP
```diff
@@ -494,18 +494,16 @@ int game_loop(const char* filename, CameraMode camera_mode) {
             }
 
             if (!Single && both_bikes_alive) {
-                if (dead1) {
-                    stop_motor_sound(true);
-                    WhoDiedFirst = 1;
-                    both_bikes_alive = false;
-                }
+                // If both die at same time, player 1 is considered to have died first
                 if (dead2) {
```
Ben3eeE: if + else if is more readable imo.

Ben3eeE (approved):
Also commit message should mention the commit that introduced the regression.

jgouly:
Can use this message maybe:

    fix(LEJATSZO): reset `WhoDiedFirst` in single player mode

    Fixes seeing 'A died first' in single player mode.

    Introduced in commit 01f3365de2250.

PR #539, opened 2026-05-12
Tidy warnings

GauChoob:
../wav.cpp:132:11: warning: method 'fade' can be made const [readability-make-member-function-const]
../wav.cpp:146:11: warning: method 'volume' can be made const [readability-make-member-function-const

seem like false positives since we assign samples[x] in the two above functions, but samples is a pointer

### DIFF
--- a/LEJATSZO.CPP
+++ b/LEJATSZO.CPP
@@ -256,10 +256,10 @@ static void physics_subframe(motorst* mot, player_keys* keys, bike_metadata* met
                 *dead = true;
             }
             if (bike_state == BikeState::Finish) {
-                *finish_time = time * TimeToCentiseconds;
+                *finish_time = (int)(time * TimeToCentiseconds);
             }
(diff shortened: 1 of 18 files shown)


Ben3eeE:
> seem like false positives since we assign samples[x] in the two above functions, but samples is a pointer

Technically const works because we are mutating the pointed to value, not the pointer. But it is a confusing API to have a const function that mutates something.

Options are:
1. Make it const, it works and is just confusing.
2. Add `NOLINTNEXTLINE(readability-make-member-function-const)` comment above the function definition.
3. Make samples an `std::vector<short>` so writing to `samples[i]` does require non-const `this`, makes clang-tidy stop complaining for the right reason.

GauChoob:
Ready for review. Brings clang-tidy warnings down from about 40 to 20, and also fixes compiler warnings about implicit type casts.

jgouly (approved):

PR #540, opened 2026-05-12
fix(chat): timestamp in local timezone

Ben3eeE:


jgouly (approved):

jgouly:
Needs windows testing

PR #557, opened 2026-05-17
Edituj part2 cleanups

jgouly:


EDITUJ.CPP
```diff
@@ -190,53 +193,57 @@ void update_and_draw_cursor() {
...
-                                "Delete Object",
-                                "Create Picture",
-                                "Delete Picture"};
+static bool RedrawEditorCanvas = false;
+void invalidate_editor_level() { RedrawEditorCanvas = true; }
+
+static bool RedrawEditorGui = false;
+void invalidate_editor_gui() { RedrawEditorGui = true; }
+
+constexpr int CommandsLength = 13;
```
Ben3eeE: wrong casing

EDITUJ.H
```diff
@@ -24,18 +24,20 @@ extern sprite* SelectedSprite;
...
+constexpr unsigned char Menu = 1;
+constexpr unsigned char MenuSelected = 2;
+constexpr unsigned char Background = 3;
+constexpr unsigned char Window = 4;
+constexpr unsigned char LevelNameWindow = 4;
+constexpr unsigned char LevelNameWindowBorder = 5;
+constexpr unsigned char WindowList = 6;
+constexpr unsigned char WindowListSelected = 7;
+constexpr unsigned char WindowButton = 8;
+constexpr unsigned char WindowInput = 9;
```
Ben3eeE: also wrong casing

jgouly: Could just make this an enum, rather than rename everything again?

EDITUJ.CPP
```diff
@@ -40,6 +40,50 @@ bool LevelChanged = false;
...
+    DeleteVertex,
+    DeletePolygon,
+    CreateFood,
+    CreateKiller,
+    DeleteObject,
+    CreateSprite,
+    DeleteSprite
+};
+
+Tool SelectedTool = Tool::Move;
```
Ben3eeE: For this commit to accurately have internal linkage shouldn't this be static and the enum class be inside an anonymous namespace?

Ben3eeE (approved):

EDITUJ.CPP
jgouly: I made that static, but left off the anon namespace.

PR #558, opened 2026-05-19
Edituj next split

jgouly:


EDITUJ.CPP
```diff
@@ -700,7 +700,7 @@ void editor() {
         }
         // Character shortcuts from text buffer
         while (has_text_input()) {
-            char c = pop_text_input();
```
Ben3eeE, line 702: This should not be using text input, but rather normal inputs and dik scancodes. It is not text input.

Ben3eeE (approved):

jgouly:
https://github.com/elmadev/elma-classic/issues/466 was an old issue for the text input change

PR #560, opened 2026-05-20
fix(sdl): drop `SDL_TEXTINPUT` events while LCtrl search is active

Ben3eeE:
On Wayland, SDL fires `SDL_TEXTINPUT` even when Ctrl is held, but only for keys whose Ctrl+key combination has no control-character mapping in xkb (0, 1, 9, comma, period, hyphen, etc.). The LCtrl-search workaround in `SDL_KEYDOWN` already inserts the character, so these keys ended up doubled in the search buffer ("1" -> "11", "11" -> "1111") while other keys worked correctly because their `SDL_TEXTINPUT` payload was a control byte and got filtered out as non-printable.

Skip `SDL_TEXTINPUT` entirely while LCtrl-search is active; the `SDL_KEYDOWN` branch is the single source of truth for that mode.

### DIFF
--- a/src/platform/sdl/sdl.cpp
+++ b/src/platform/sdl/sdl.cpp
@@ -433,8 +433,9 @@ void handle_events() {
             }
 #endif
 
-            // SDL doesn't generate text input events when Ctrl is held
-            // Resolve layout-specific keycodes to support LCtrl search
+            // On macos, SDL suppresses SDL_TEXTINPUT while Ctrl is held, so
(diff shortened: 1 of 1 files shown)


GauChoob:
I don't notice any difference pre and post commit on windows.

My professional opinion is we should drop Left Control search completely :P

Anyways, if joey approves these feel free to merge it

jgouly (approved):

PR #561, opened 2026-05-20
fix(input): keep numpad as numpad during gameplay

Ben3eeE:
The NumLock-off numpad-to-arrow remap in `keyboard::end_frame()` runs globally, so any gameplay binding on numpad-6 also fires the Right Arrow binding (and likewise for 8/4/2/9/3/period). Multiplayer setups that overload the right cluster (e.g. disconnect on Right Arrow, right-volt on numpad-6) trigger both actions on a single keypress, and users have to toggle NumLock every time they switch between menus and play.

Add a `keyboard::set_numpad_nav` toggle and disable the remap for the duration of game_loop, so during play scancode in = scancode out. Other places keep respecting numlock to prevent firing both textinput and navigation keys.

### DIFF
--- a/LEJATSZO.CPP
+++ b/LEJATSZO.CPP
@@ -14,6 +14,7 @@
 #include "object.h"
 #include "physics_init.h"
 #include "platform/implementation.h"
+#include "platform/sdl/keyboard.h"
 #include "segments.h"
 #include "timer.h"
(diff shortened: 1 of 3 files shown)


GauChoob (approved):
Works for me

PR #562, opened 2026-05-20
feat(replay): default merged replays to single view

Ben3eeE:


jgouly (approved):

recorder.cpp
```diff
@@ -15,6 +15,7 @@
 recorder* Rec1 = nullptr;
 recorder* Rec2 = nullptr;
 int MultiplayerRec = 0;
+int MergedRec = 0;
```
GauChoob: As a bool please

GauChoob (approved):
Please make it a bool and then can merge after

PR #565, opened 2026-05-21
Editor input buffer bug

GauChoob:
Fixes #466


jgouly (approved):
I think the `i == 0` `i-- == 0` is a bit of a strange pattern, approving anyway maybe bene will merge.

PR #566, opened 2026-05-22
CLang-tidy cleanup

GauChoob:
We drop the "const" clang-tidy check because we can always trivially add it later if we have a const object, and it doesn't play nice with the wav class

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,7 +32,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.x'
+          python-version: "3.x"
 
       - name: Install dependencies
         run: python -m pip install meson==1.10.0 ninja
@@ -57,7 +57,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.x'
+          python-version: "3.x"
 
       - name: Install dependencies
         run: python -m pip install meson==1.10.0 ninja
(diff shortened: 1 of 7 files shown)


GauChoob:
Can drop the last commit when approved or reviewed, but currently we demonstrate both a passing and a failing workflow

jgouly:
I think we should just mark that wav function as const, better to have that lint than drop it for a 'corner case'.

GauChoob:
I could refactor wav to use a vector instead of a pointer which should solve the problem, was just too lazy to touch wav since unlikely it needs to ever be touched again

jgouly:
Just leave it as an array and mark it const and be done with it :P

Ben3eeE:
The TreatWarningsAsErrors change and removing show summary is not the way to go. It gives us less information in the pipeline output which forces us to download the txt file to see results. Also currently errors are when clang-tidy failed to run completely and a warning is clang-tidy found style issues with the code keeping them separated makes sense for this reason.

I would instead retain the summary step and exit non zero on warnings:
```diff
diff --git a/.clang-tidy b/.clang-tidy
index 19a64e88..fc88a654 100644
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -31,6 +31,6 @@ Checks: >
   readability-static-accessed-through-instance,
   readability-use-std-min-max,
   readability-string-compare
-WarningsAsErrors: "*"
+WarningsAsErrors: ""
 HeaderFilterRegex: '^(?!.*(subprojects|include/glad|include/directinput/scancodes\.h|include/sdl/scancodes_windows\.h)).*'
 FormatStyle: file
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index cb0b9e18..3caadbaa 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,18 +68,25 @@ jobs:
       - name: Run clang-tidy
         run: |
           ninja -C build clang-tidy > clang-tidy-output.txt 2>&1 || true
-          if grep -qE "error:" clang-tidy-output.txt; then
+          if grep -qE "warning:|error:" clang-tidy-output.txt; then
             echo "Clang-tidy found issues."
-            echo ""
-            grep -E "warning:|error:" clang-tidy-output.txt 2>/dev/null || true
-            echo ""
-            echo "Full output available in artifacts"
             exit 1
           fi
 
           echo "No clang-tidy issues found."
           exit 0
 
+      - name: Show summary
+        if: always()
+        run: |
+          echo "=== Clang-tidy Summary ==="
+          grep -E "warning:|error:" clang-tidy-output.txt 2>/dev/null || true
+          echo ""
+          echo "Total warnings: $(grep -c "warning:" clang-tidy-output.txt || true)"
+          echo "Total errors: $(grep -c "error:" clang-tidy-output.txt || true)"
+          echo ""
+          echo "Full output available in artifacts"
+
       - name: Upload clang-tidy results
         if: always()
         uses: actions/upload-artifact@v7
```

Ben3eeE:
Also merging this will make the private side pipeline fail, so I would want to see a PR on private side to clean up those things.

GauChoob:
We can keep the summary/your recommended changes but also exit 1 on the summary step instead of the previous step. I will do that change after work

Ben3eeE:
This doesn't exit with a non-zero exit code if there are warnings like the diff I provided does?

jgouly (approved):

PR #574, opened 2026-05-30
Skip intro

jgouly:


eol_settings.h
```diff
@@ -113,6 +113,7 @@ class eol_settings {
     Default<bool> show_demo_menu_{true};
     Default<bool> show_help_menu_{true};
     Default<bool> show_best_times_menu_{true};
+    Default<bool> skip_intro_{true};
```
GauChoob: Should be false by default

src/menu/intro.cpp
```diff
@@ -73,20 +75,22 @@ void menu_intro() {
...
-        show_intro_screen();
     }
 
     if (State->player_count == 0) {
         if (!menu_player_create(true)) {
             menu_exit();
         }
     } else {
-        if (!menu_player_choose(true, false)) {
+        if (!EolSettings->skip_intro() && !menu_player_choose(true, false)) {
```
GauChoob: Using && for logic seems a bit unobvious. Maybe split into two if statements instead or something?

GauChoob:
[diff.txt](https://github.com/user-attachments/files/28430317/diff.txt)

Here's my suggestion to make the intro animation more simple instead of more complicated

GauChoob (changes requested):

GauChoob (approved):

PR #578, opened 2026-05-31
Windowed editor

### LINKED ISSUE #576: In Editor, mouse shape (+ or X) is frozen when a subwindow is open
Should probably always be +

GauChoob:
Fixes https://github.com/elmadev/elma-classic/issues/576
Fixes https://github.com/elmadev/elma-classic/issues/467

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -248,19 +248,6 @@ void draw_cursor() {
     draw_cursor_shape();
 }
 
-void update_and_draw_cursor() {
-    int mouse_x = 0;
-    int mouse_y = 0;
-    get_mouse_position(&mouse_x, &mouse_y);
-    if (mouse_x == MouseX && mouse_y == MouseY) {
-        return;
-    }
-    erase_cursor();
-    MouseX = mouse_x;
-    MouseY = mouse_y;
-    draw_cursor();
-}
-
 void draw_tooltip(const char* text) {
     static char TooltipText[110] = "";
     if (text) {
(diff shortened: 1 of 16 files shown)


src/editor/editor.cpp
```diff
@@ -171,66 +171,32 @@ int MouseX = 0;
 int MouseY = 0;
 static bool CursorShapeIsX = false;
 
-static void draw_cursor_pixel(int x, int y) {
-    if (x < 0 || y < 0 || x > (SCREEN_WIDTH - 1) || y > (SCREEN_HEIGHT - 1)) {
+static void draw_cursor_pixel(pic8& dest, int x, int y) {
+    if (x < 0 || y < 0 || x >= dest.get_width() || y > dest.get_height()) {
```
Ben3eeE: Why `x >= dest.get_width()` and `y > dest.get_height()`?

Ben3eeE:
When creating a new polygon with the Create Vertex tool, the cursor now shifts to "X" when within grab distance of an existing vertex. This is because the mousemove refactor dropped the continue statements, so draw_mouse_tooltip() (which recomputes CursorShapeIsX from the closest entity) now runs on every move during polygon creation.

This is a difference in behavior to Elma.

src/editor/editor.cpp
GauChoob: Fixed; changed both to >=

GauChoob:
> When creating a new polygon with the Create Vertex tool, the cursor now shifts to "X" when within grab distance of an existing vertex. This is because the mousemove refactor dropped the continue statements, so draw_mouse_tooltip() (which recomputes CursorShapeIsX from the closest entity) now runs on every move during polygon creation.
> 
> This is a difference in behavior to Elma.

Fixed! (turned it into an else if chain instead of an if chain)

jgouly (approved):

PR #583, opened 2026-06-03
Lejatszo Cleanup

GauChoob:
Instead of having the hardcoded vars Motor1, metadata1, Rec1, HudVisibility1, dead1, finish_time1, State->keys1 and passing up to 7 arguments to sub functions, wrap this set of vars into a class called `driver` for simplicity

### DIFF
--- a/include/game/driver.h
+++ b/include/game/driver.h
@@ -3,22 +3,43 @@
 
 #include "game/recorder.h"
 
-struct bike_metadata;
 struct hud_visibility;
 struct motorst;
 struct player_keys;
 
+struct turning_data {
+    int flipped;
+    double turn_time;
+    double turn_phase;
+};
+
+struct bike_metadata {
+    bike_sound sound;
+    double volt_time;
+    bool volt_is_right;
+
+    bool draw_view;
+
+    bool turn_key_previous;
+    bool one_turn_used;
+
+    double arm_position;
+
+    turning_data bike_turning;
+    turning_data camera_turning;
+};
+
 struct driver {
     motorst* mot;
-    bike_metadata* meta;
+    bike_metadata meta;
     bool dead = false;
     int finish_time = 0;
     recorder* rec;
     player_keys* keys;
     hud_visibility* hud;
 
-    driver(motorst* mot, bike_metadata* meta, recorder* rec, player_keys* keys,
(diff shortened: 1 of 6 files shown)


src/game/game.cpp
```diff
@@ -311,8 +317,10 @@ static void update_camera_turn_phase(turning_data* data, double time, int flippe
     }
 }
 
-static void update_graphical_metadata(motorst& mot, bike_metadata& metadata, recorder* rec,
-                                      double time) {
+static void update_graphical_metadata(driver& driv, recorder* rec, double time) {
```
Ben3eeE, line 320: Doesn't driver have a recorder? Why are there two?

include/game/driver.h
```diff
@@ -0,0 +1,45 @@
...
+    bool turn_key_previous;
+    bool one_turn_used;
+
+    double arm_position;
+
+    turning_data bike_turning;
+    turning_data camera_turning;
+};
+
+class driver {
```
Ben3eeE: Since this is all public members should it just be a struct and not a class?

Ben3eeE:
What is the future plan? Where will be put something like `max_speed` that we will need for speedometer and speed battles?

src/game/game.cpp
GauChoob: For this function, `recorder* rec` actually means `bool is_game`, i.e. a bool to check if the parent function is game_loop() vs replay_loop(). replay_loop() passes nullptr and game_loop() passes rec. To keep this PR simple, I considered it beyond the scope of this current PR (among a bunch of other refactors that could be done to improve/simplify the code if we add the driver class)

So in other words, this param will hopefully become a bool in a future PR, but currently it needs to stay separate

include/game/driver.h
GauChoob: My C knowledge and lack of C++ knowledge is showing here. At the moment it's used as a struct, though in my head it might in the future gain functionality that makes it more like a class. I don't really care either way, it can be a struct

GauChoob:
Future plan yes is to put things like max_speed, bug_bounce_data, and other stuff directly into driver instead of randomly appending variables to bike_metadata or motorst

driver should contain all the data that is specific to one bike

Ben3eeE (approved):

PR #587, opened 2026-06-04
feat(editor/canvas): allow zooming away from level

jgouly:
I split this out cos I wasn't sure about it

### DIFF
--- a/src/editor/canvas.cpp
+++ b/src/editor/canvas.cpp
@@ -15,11 +15,9 @@ constexpr double ZOOM_IN_LIMIT = 0.017;
 constexpr double ZOOM_OUT_LIMIT = 500.0;
 constexpr double ZOOM_FACTOR = 170.0;
 static double AspectRatio = 1.0;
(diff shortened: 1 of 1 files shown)


Ben3eeE:
I have no idea what this changes. Can you explain?

GauChoob:
This part of the code prevents you from panning too far away from the level
image

This way, it's impossible to make a level that crashes due to being too large because you physically cannot see out of bounds and so you can't click to create a vertex that would be out of bounds
image

Hib increased the max zoom out for convenience, breaking the above assumption and allowing you to create a level that is too large, crashing the game. That's why he had to add a new level size topology check.

So the original purpose of this code doesn't exist anymore, so it's reasonable to delete it. As a side effect, you can slowly zoom in/out to shift your viewpoint away from the level itself if you really want to

jgouly:
Did EOL actually remove this code (as it says in the commit message)? Other than that seems good

GauChoob:
This code is commented out in asm by either hib or mila

Ben3eeE (approved):
Makes sense to me

PR #588, opened 2026-06-04
Move4 p2

jgouly:
The keys.cpp scandal


Ben3eeE:
We have namespace util::text which is something completely different. It can't be util/text.cpp

GauChoob:
Let me know what you want and I'll put it in the next batch of moves. I don't care that much though keys seemed wrong to me. Input.cpp?

PR #602, opened 2026-06-15
FPS limiter

### LINKED ISSUE #507: One frame brake should be conserved between physics frames and reset in the outer loop
This is mandatory to fix before any public release, allowing manipulating input between different physics frames in the inner loop is a very controversial change that makes pops even more overpowered. In EOL the input is fixed in the outer loop.

Ben3eeE:
Fixes https://github.com/elmadev/elma-classic/issues/507

Faithful reimplementation of eol fps limiter. Was briefly tested by Zamppe and FM to ensure it works great (Before rebasing)

Still WIP.

Undecided:
* Should we allow decimals in configured FPS value?

### DIFF
--- a/include/eol/pacer.h
+++ b/include/eol/pacer.h
@@ -1,0 +1,18 @@
+#ifndef EOL_PACER_H
+#define EOL_PACER_H
+
+namespace pacer {
+
+constexpr double PHYS_MAX_DELTA = 0.0055;
+
+void reset();
+
+// Tell the pacer that wall-clock time has advanced to `cel` cel-units. Used as
+// the catch-up target by phys_step_next.
+void set_time_passed(double cel);
+
+bool phys_step_next(double* out_dt);
+
+} // namespace pacer
+
+#endif
--- a/include/eol/pacer.h
+++ b/include/eol/pacer.h
@@ -7,9 +7,7 @@ constexpr double PHYS_MAX_DELTA = 0.0055;
 
 void reset();
 
-// Tell the pacer that wall-clock time has advanced to `cel` cel-units. Used as
-// the catch-up target by phys_step_next.
-void set_time_passed(double cel);
+void tick();
 
 bool phys_step_next(double* out_dt);
 
--- a/src/eol/fps.cpp
+++ b/src/eol/fps.cpp
@@ -1,0 +1,112 @@
+#include "eol/fps.h"
+#include "eol/settings.h"
+#include "eol/status_messages.h"
+#include <cmath>
+#include <format>
+
+int CurrentFps = 0;
+bool LiveFpsLimitEnabled = false;
+float LiveFpsLimit = 0.0f;
+
+namespace fps {
+
+namespace {
+
+// Shorter window for the first update to get a quicker initial reading
+constexpr int COUNTER_INITIAL_INTERVAL_MS = 100;
+// After first update use a longer interval for more stable readings
+constexpr int COUNTER_UPDATE_INTERVAL_MS = 1000;
+
+int recent_frames = 0;
+long long counter_window_ms = 0;
+bool counter_initial = true;
+
+} // namespace
+
+void reset_counter(long long now_ms) {
+    recent_frames = 0;
+    counter_window_ms = now_ms;
+    counter_initial = true;
+    CurrentFps = 0;
+}
+
+void tick_counter(long long now_ms) {
+    recent_frames++;
+
+    if (!EolSettings->show_fps_info()) {
+        return;
+    }
+
+    const long long window_target =
+        counter_initial ? COUNTER_INITIAL_INTERVAL_MS : COUNTER_UPDATE_INTERVAL_MS;
+
+    const long long window_elapsed = now_ms - counter_window_ms;
+
+    if (window_elapsed >= window_target) {
+        CurrentFps = (int)std::round((double)recent_frames / (double)window_elapsed * 1000.0);
+
+        if (counter_initial) {
+            counter_initial = false;
+        } else {
+            counter_window_ms = now_ms;
+            recent_frames = 0;
+        }
+    }
+}
+
+// Two distinct states are at play here:
+//   - Live: what the pacer is currently using (LiveFpsLimit / LiveFpsLimitEnabled).
+//     Doesn't change mid-run.
+//   - Pending: what EolSettings holds. Applied to Live at the next run start.
+//
+// Each public operation reduces to: compute a wanted Pending state, apply it,
+// then tell the user whether it changes Live or matches it.
+namespace {
+
+struct fps_pending {
+    float limit;
+    bool enabled;
+};
+
+void set_pending(fps_pending p) {
+    EolSettings->set_fps_limit(p.limit);
+    EolSettings->set_fps_limit_enabled(p.enabled);
+}
+
+bool live_matches(fps_pending p) {
+    if (!p.enabled) {
+        return !LiveFpsLimitEnabled;
+    }
+    return LiveFpsLimitEnabled && LiveFpsLimit == p.limit;
+}
+
+void request(fps_pending wanted) {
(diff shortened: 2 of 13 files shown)


GauChoob:
I will check this exciting PR later but I think we should disallow non-integer FPS limit because we want people to optimize their gameplay, and not over-optimize settings

Ben3eeE:
I don't have any time to do anything about until after fem has ended but wanted to open so it can be included in FEM release for more testing.

Float values are there for additional testing because a even faithful reimplementation messes for internalists but undecided what to eventually do. All that is decided now is that I have checked with Markku that it would be possible to do without major changes. A decision needs to be made eventually.

include/eol/pacer.h
```diff
@@ -0,0 +1,18 @@
+#ifndef EOL_PACER_H
```
GauChoob, line 1: Imo the pacer runs regardless if online, so it doesn't need to be in eol/ folder per se

include/eol/pacer.h
```diff
@@ -0,0 +1,18 @@
...
+#define EOL_PACER_H
```
GauChoob, line 2: Commits and some vars need to be translated

src/eol/settings.cpp
```diff
@@ -177,6 +178,12 @@ void eol_settings::set_show_gravity_arrows(bool b) { show_gravity_arrows_ = b; }
 
 void eol_settings::set_recording_fps(int fps) { recording_fps_ = fps; }
 
+void eol_settings::set_fps_limit_enabled(bool b) { fps_limit_enabled_ = b; }
+
+void eol_settings::set_fps_limit(float v) { fps_limit_ = v; }
```
GauChoob, line 183: Why not just use double here like everywhere else? Imo better to be simple and use double everywhere unless we have a specific reason to use float

src/eol/fps.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "eol/settings.h"
+#include "eol/status_messages.h"
+#include <cmath>
+#include <format>
+
+int CurrentFps = 0;
+bool LiveFpsLimitEnabled = false;
+float LiveFpsLimit = 0.0f;
+
+namespace fps {
```
GauChoob, line 11: I prefer my implementation of namespace fps here <https://github.com/elmadev/elma-classic/pull/600/changes/e08ead5bd48676f6d919f254612288904ea91ee7> but both seem good. Mine keeps track of both UPS and FPS which imo is worthwhile (physics frames and render frames)

src/eol/fps.cpp
```diff
@@ -0,0 +1,112 @@
...
+// Each public operation reduces to: compute a wanted Pending state, apply it,
+// then tell the user whether it changes Live or matches it.
+namespace {
+
+struct fps_pending {
+    float limit;
+    bool enabled;
+};
+
+void set_pending(fps_pending p) {
```
GauChoob, line 71: For some reason I find all the function names in this file really hard to intuitively understand. They all seem a little bit vague

Some ideas:

```
bool LiveFpsLimitEnabled; float LiveFpsLimit; -> use the struct fps_pending
rename struct fps_pending -> fps_limit;
rename void set_pending() -> conceptually, eol settings is supposed to call the the namespace function, and not the other way around, so this should be deleted.
rename bool live_matches() -> fps_limit operator==
rename void request() -> set_next_fps_limit
```

src/eol/fps.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Two distinct states are at play here:
+//   - Live: what the pacer is currently using (LiveFpsLimit / LiveFpsLimitEnabled).
+//     Doesn't change mid-run.
+//   - Pending: what EolSettings holds. Applied to Live at the next run start.
+//
+// Each public operation reduces to: compute a wanted Pending state, apply it,
+// then tell the user whether it changes Live or matches it.
+namespace {
```
GauChoob, line 64: Imo all this fps setting stuff should go directly into the pacer

src/eol/pacer.cpp
```diff
@@ -1,16 +1,66 @@
...
+// eol-client-style time-freeze limiter.
+static long long start_time_ms = 0;
+static long long last_real_frame_ms = 0;
+static uint64_t total_real_frames = 0;
+
+} // namespace
+
 void reset() {
+    long long now = get_milliseconds();
+    LiveFpsLimitEnabled = EolSettings->fps_limit_enabled();
```
GauChoob, line 28: If fps setting stuff is moved from fps to pacer, LiveFpsLimitEnabled and LiveFpsLimit can be managed inside the translation unit using the struct that is also used internally

src/eol/pacer.cpp
```diff
@@ -1,16 +1,66 @@
...
 }
 
-void set_time_passed(double cel) { time_passed_ = cel; }
+void tick() {
+    long long now_ms = get_milliseconds();
+
+    if (LiveFpsLimitEnabled) {
+        long long elapsed = now_ms - start_time_ms;
+        long long since_last = now_ms - last_real_frame_ms;
+        bool over_limit =
```
GauChoob, line 47: This line is really hard to understand. Maybe break it into two parts: `long long total_allowed_frames = elapsed*LiveFpsLimit // multiplied by a factor of 1000 as elapsed is in milliseconds`

GauChoob: And add a comment explaining that here is where we are actually throttling the fps

src/eol/pacer.cpp
```diff
@@ -1,16 +1,66 @@
...
+            (total_real_frames * 1000ULL > (uint64_t)elapsed * (uint64_t)LiveFpsLimit);
+        if (over_limit && since_last <= LIMITER_TIMEOUT_MS) {
```
GauChoob, line 49: Conceptually, is the LIMITER_TIMEOUT_MS just mimicking an FPS cap of >= 30.303030?

If so imo would be better to simplify by doing std::max(LiveFpsLimit, 30.303030) instead of having this extra step which is hard to understand

src/eol/settings.cpp
```diff
@@ -177,6 +178,12 @@ void eol_settings::set_show_gravity_arrows(bool b) { show_gravity_arrows_ = b; }
 
 void eol_settings::set_recording_fps(int fps) { recording_fps_ = fps; }
 
+void eol_settings::set_fps_limit_enabled(bool b) { fps_limit_enabled_ = b; }
```
GauChoob, line 181: Conceptually, the eol_settings::setter should at this point call the fps namespace and/or pacer namespace, and not the other way around

src/renderer/render.cpp
```diff
@@ -607,6 +609,37 @@ static bool bike_in_view(const motorst* mot, vect2 center) {
...
 }
 
+struct info_panel_row {
+    std::string label;
+    std::string value;
+};
+
+// Render the bottom-right info panel: rows[0] is the bottom row, each later row stacks above it
+// (the backbuffer is upside-down, so larger y is higher on screen).
+static void render_info_panel(pic8* pic, const std::vector<info_panel_row>& rows) {
```
GauChoob, line 619: Conceptually this is great

src/game/game.cpp
```diff
@@ -141,7 +151,9 @@ static void physics_subframe(driver& driv, double time, double dt) {
     // Adjust key inputs to only allow valid inputs, accounting for volt delay and cripples
     bool is_gas_down = is_game_key_down(keys->gas);
     bool is_brake_down = is_game_key_down(keys->brake) || is_game_key_down(keys->brake_alias) ||
-                         was_game_key_just_pressed(keys->one_frame_brake);
+                         was_game_key_just_pressed(keys->one_frame_brake) ||
```
GauChoob, line 154: Can we refactor this to only use the latch and not was_game_key_just_pressed(keys->one_frame_brake)?

GauChoob (commented):
Here are some of my thoughts. Overall seems pretty cool

PR #605, opened 2026-06-22
refactor(render): lay out the corner info panel as a table

GauChoob:
Copies current spacing formatting of corner info panel


Ben3eeE (approved):
I approve because it's my own commit

GauChoob:
I also approve because it's bene's commit

PR #608, opened 2026-06-22
Replay last apple time

### LINKED ISSUE #580: Last apple time is bugged when rewinding in replay
Time doesn't update

Ben3eeE:
- First commit fixes [elmadev/elma-classic#580](https://github.com/elmadev/elma-classic/issues/580).
- Second commit refactors `recall_event` to return the event, enabling the third commit.
- Third commit uses the recorded event time so the last apple time is shown accurately (and deterministically) in replays, previously it flickered between e.g. 4:37 and 4:38.
- Fourth commit is opinionated: `recorder::gravity` differing from its sibling functions for no apparent reason is confusing, so this aligns it. Also the name `gravity` is quite bad imo but I didn't change it, could also be aligned like `find_last_gravity`.

### DIFF
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -671,8 +671,7 @@ static void rewind_override_animations(driver& driv, double time) {
     }
     metadata->camera_turning.flipped = flipped_camera;
 
-    metadata->volt_time =
-        rec->find_last_volt_time(time, &metadata->volt_is_right).value_or(-1000.0);
+    metadata->volt_time = rec->last_volt_time(&metadata->volt_is_right).value_or(-1000.0);
 }
 
 // Load replay data (instead of simulating bike physics)
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -646,6 +646,8 @@ static void reverse_events(driver& driv, double time) {
             obj->active = true;
             mot->apple_count--;
 
+            mot->last_apple_time = (int)(rec->last_apple_time().value_or(0) * TIME_TO_CENTISECONDS);
+
             if (obj->gravity()) {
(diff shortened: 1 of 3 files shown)


src/game/game.cpp
```diff
@@ -645,6 +645,7 @@ static void reverse_events(driver& driv, double time) {
         if (obj && obj->type == object::Type::Food) {
             obj->active = true;
             mot->apple_count--;
+            mot->last_apple_time = (int)(rec->find_last_apple_time(time) * TIME_TO_CENTISECONDS);
```
Ben3eeE: This technically does `-1000 * TIME_TO_CENTISECONDS` when rewinding past grabbing first apple. But last apple time is not shown when no apples are taken so it doesn't matter what value `mot->last_apple_time` gets.

GauChoob: Imo should std::optional<> the value, especially since we aren't really concerned about overhead here, or worst case return 0 instead of using magic number

src/game/recorder.cpp
```diff
@@ -374,6 +374,18 @@ double recorder::find_last_volt_time(double time, bool* is_right_volt) const {
     return -1000.0;
 }
 
+double recorder::find_last_apple_time(double time) const {
```
GauChoob: Why do we need to specify a time? Can't we just imitate the same behaviour of gravity() and rewind it to the previous apple automatically without specifying the current time?

GauChoob (commented):
I feel like find_last_apple_time() doesn't need the time param but I didn't test

GauChoob:
Idm find_last_gravity btw, you can add it to last commit if you want

GauChoob (approved):
Looks good to me

PR #610, opened 2026-06-23
Fps/Ups combined PR

GauChoob:
Here's kind of what I had in mind as a combined PR

I added UPS boost to the end of this PR just to show how it would fit in, but it wouldn't be part of the first merge

My notes for bene:

I changed the FPS limiter to use an int. We can also change it to float/double in the future, but we can never go back. So keep it an int until we decide we want it floating-point

bene's commit order is changed slightly

UPS/FPS counter and the UPS pacer are separated into two distinct files. The UPS/FPS counter is based on my PR. pacer is based on bene's PR (file moved out from eol/ to game/pacer.cpp)

Instead of using LiveFps, wanted-fps and EolSettings->fps, simplify by only using LiveFps and EolSettings->fps

Slightly change the FPS status message behaviour (change from old client):

- Even if fps info is hidden from bottom-right, still send a status message
- Even if you request the same !fps twice, still send a status message
- - If the requested fps change is identical to both the current run's fps AND the currently requested fps, tell the user he is dumb, otherwise tell the user that there will be a new fps next run

I did some cosmetic changes to the fps limiter:
- renamed some things as they made sense to me, but we can change some variable names back to benanames if you don't like the new names.
- I used long long everywhere instead of mix of uint64_t and long long

### DIFF
--- a/src/renderer/render.cpp
+++ b/src/renderer/render.cpp
@@ -853,7 +853,7 @@ static void render_view(bool player1, bool bottom_player, pic8* pic, double time
 }
 
 void render_game(double time, driver& driv1, driver& driv2, camera& current_camera, GameLoop loop) {
-    fps::count_fps();
+    fps::count_gpu();
 
     // Determine who we are going to draw (player 1, player 2 or both)
     bool draw_player1 = driv1.draw_view;
--- a/src/renderer/render.cpp
+++ b/src/renderer/render.cpp
@@ -12,6 +12,7 @@
 #include "main.h"
 #include "physics/flagtag.h"
 #include "physics/init.h"
+#include "physics/pacer.h"
 #include "pic/abc8.h"
 #include "pic/anim.h"
 #include "pic/lgr.h"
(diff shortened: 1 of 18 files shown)


src/renderer/render.cpp
```diff
@@ -824,6 +826,112 @@ static void render_view(bool player1, pic8* pic, double time, driver& driv, driv
...
+                    if (limit != next_limit) {
+                        ups_text = std::format("{} ({} -> {})", ups_text, format_limit(limit),
+                                               format_limit(next_limit));
+                    } else {
+                        ups_text = std::format("{} ({})", ups_text, format_limit(limit));
+                    }
+                }
+            }
+
+            info_rows.push_back({"UPS", std::move(ups_text)});
```
GauChoob: UPS rows and FPS rows have duplicate code that can probably be merged together

GauChoob:
LIMITER_TIMEOUT deleted as it was a temporary fix that is no longer needed. This change should only impact behaviour if you get two big lag spikes in a really short amount of time.

PR #612, opened 2026-07-13
feat(editor): display filename as-is

jgouly:


GauChoob:
What issue is this solving? Also if we do this change, then we can delete strupr from platform utils

GauChoob (approved):

PR #624, opened 2026-07-24
Default Ground/Sky

GauChoob:


Ben3eeE:
People have requested it be split into two settings so can default sky and leave ground what it is.

GauChoob:
Split ground and sky options to be separate

jgouly (approved):

src/eol/settings.cpp
```diff
@@ -521,7 +524,10 @@ void eol_settings::sync_controls_from_state(state* s) {
     EolSettings->persist_brake_alias_key_player_b(s->keys2.brake_alias);
     EolSettings->persist_one_frame_brake_key_player_a(s->keys1.one_frame_brake);
     EolSettings->persist_one_frame_brake_key_player_b(s->keys2.one_frame_brake);
+
     EolSettings->persist_escape_alias_key(s->key_escape_alias);
+    EolSettings->persist_default_ground_sky_key(s->key_default_ground_sky);
```
Ben3eeE, line 540: I don't understand these newline changes in 065590d249ae2780fdef01dd5f11f1a7f00f5e6a it doesn't make sense to have these two (esc alias and default ground sky) in the same block/group.

src/menu/options.cpp
```diff
@@ -273,6 +273,9 @@ void menu_options() {
                 invalidate_level();
             });
 
+        BOOL_OPTION("Override Ground Texture:", default_ground);
```
Ben3eeE: Text is too long for menu (with 1024x768 resolution)

src/game/game.cpp
```diff
@@ -399,6 +399,31 @@ static void handle_eol_inputs() {
     if (was_game_key_just_pressed(State->key_battle_leader)) {
         EolClient->toggle_show_battle_leader();
     }
+
+    if (was_game_key_just_pressed(State->key_default_ground_sky)) {
+        unsigned int permutation =
+            (EolSettings->default_ground() << 1) | EolSettings->default_sky();
+        permutation = (permutation - 1) % 4;
```
Ben3eeE: `(permutation - 1) % 4` relies on unsigned wraparound to map 0 to 3. `(permutation + 3) % 4` says the same thing without the trick

Or if we want it in growing order which feels more natural it could be `(permutation + 1) % 4`

src/game/game.cpp
```diff
@@ -399,6 +399,31 @@ static void handle_eol_inputs() {
...
+        bool ground = (bool)(permutation & 2);
+        bool sky = (bool)(permutation & 1);
```
Ben3eeE: Why c style casts instead of `bool ground = (permutation & 2) != 0; `?

src/pic/lgr.cpp
```diff
@@ -1043,19 +1043,19 @@ lgrfile::~lgrfile() {
     }
 }
 
-void lgrfile::sanitize_default_texture_names() {
+void lgrfile::sanitize_default_texture_names(char foreground_name[10], char background_name[10]) {
```
Ben3eeE: Parameter names shadow lgrfile members of the same names. Also in `reload_default_textures` it becomes confusing.

src/pic/lgr.cpp
```diff
@@ -1043,6 +1043,53 @@ lgrfile::~lgrfile() {
     }
 }
 
+void lgrfile::sanitize_default_texture_names() {
```
Ben3eeE: a name collision always erases the background, so with "Override Sky" on and a level whose foreground is literally sky, the override silently becomes a random fallback texture. the override should probably win collisions

Ben3eeE (dismissed):
Why so many commits for adding an option? When I added cripples I got complaints there were multiple commits for basically what the first three commits do and I should squash it into one because the menu is now just a single line. What is it we are doing? Is it to squash this to single commit or not?

src/menu/options.cpp
jgouly: Should probably call it default, that's what the current option is, assuming it fits.

src/game/game.cpp
Ben3eeE: Could just be:

```c++
        const unsigned int permutation =
            ((EolSettings->default_ground() << 1 | EolSettings->default_sky()) + 3) % 4;
        const bool ground = (permutation & 2) != 0;
        const bool sky = (permutation & 1) != 0;
```

src/eol/settings.cpp
GauChoob: Esc alias and default are currently categorized as universal controls (at least for now until we have more keys), so in controls.cpp they are lumped together. But it's mainly to separate them out from the keys above

GauChoob:
> Why so many commits for adding an option? When I added cripples I got complaints there were multiple commits for basically what the first three commits do and I should squash it into one because the menu is now just a single line. What is it we are doing? Is it to squash this to single commit or not?

Imo it doesn't matter either way. I'm happy to have it squashed a bit. In this specific case it was separated because there's some finicky synchronization stuff to do with controls.cpp and also the keybind implementation with permutations is also a bit special

So tldr whichever way is fine by me

src/menu/options.cpp
GauChoob: I disliked "default" because we are actually overriding the default foreground/background texture specified in the level. But I guess all ways of wording have advantages and disadvantages.

Anyways open to other wording

src/game/game.cpp
GauChoob: Looks good. +3 is better than -1. I feel decrementing is better than incrementing because that way it goes from no override -> double override, making the effect more obvious 

I would probably split +3 % 4 into own line just for clarity more than anything though

src/pic/lgr.cpp
GauChoob: Yes we can do that

src/eol/settings.cpp
GauChoob: I looked at code, and imo the separation makes sense. We have 4 sections:
- player keys (alovolt, brake alias, one frame brake)
- universal keys (esc alias, default ground/sky key)
- replay keys
- function keys (F1, F2, etc)

So I did not change anything. We can always re-organize in the future when we have even more keybinds

src/menu/options.cpp
GauChoob: I changed it to "Override Ground" and "Override Sky" instead of "Override Ground Texture". If someone feels strongly about calling it "default" we can change it to that

GauChoob:
I changed the permutation code a bit from joey's because of type conversion warnings in MSVC

GauChoob:
Sqaushed the first two commits together at least

include/pic/lgr.h
```diff
@@ -75,6 +76,8 @@ class lgrfile {
     void add_texture(pic8* pic, piclist* list, int index);
     void add_mask(pic8* pic, piclist* list, int index);
 
+    void sanitize_default_texture_names(char foreground[10], char background[10]);
```
Ben3eeE: This is still shadowing lgrfile class members even after the rename, just different members

changelog.txt
```diff
@@ -60,6 +60,7 @@
 * Editor cursor no longer disappears when inputting a clipping or distance
 * When saving internal replay, internal number and name will be shown instead of QWQUU0XX.lev
 * Flowers can be created/deleted in the internal editor.
+* Default foreground/background textures no longer get reset if texture not found or wrong lgr used
```
Ben3eeE, line 63: Also missing changelog for new default keybinding and split options.

PR #625, opened 2026-07-26
ASSERT_INTERNAL

### LINKED ISSUE #621: internal_assert
```cpp
   assert_internal(Lgr->texture_count < 2)
```

GauChoob:
I added it to dialog.cpp as an example even though it's kind of self-explanatory

Resolves https://github.com/elmadev/elma-classic/issues/621

### DIFF
--- a/include/main.h
+++ b/include/main.h
@@ -31,4 +31,11 @@ void delay(int milliseconds);
 [[noreturn]] void external_error(const std::string& message,
                                  std::source_location loc = std::source_location::current());
 
+#define ELMA_ASSERT(condition)                                                                     \
+    do {                                                                                           \
+        if (!(condition)) {                                                                        \
+            internal_error("Assertion failed: " #condition);                                       \
+        }                                                                                          \
+    } while (0)
+
 #endif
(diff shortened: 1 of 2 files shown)


include/main.h
```diff
@@ -31,4 +31,9 @@ void delay(int milliseconds);
 [[noreturn]] void external_error(const std::string& message,
                                  std::source_location loc = std::source_location::current());
 
+#define ASSERT_INTERNAL(condition)                                                                 \
```
Ben3eeE: Change to this:

```c++
#define ASSERT_INTERNAL(condition)                                                                 \
      do {                                                                                           \
          if (!(condition)) {                                                                        \
              internal_error("Assertion failed: " #condition);                                       \
          }                                                                                          \
      } while (0)
```

Ben3eeE (approved):

Ben3eeE:
LGTM

jgouly:
what about ELMA_ASSERT or EOL_ASSERT? Or just ASSERT?

GauChoob:
Either ASSERT_INTERNAL, INTERNAL_ASSERT and ELMA_ASSERT all seem good to me. It's technically not a global assert in the sense that it throws an internal error (so you shouldn't use it when reading external files for example)

Shall we go with ELMA_ASSERT?

PR #627, opened 2026-07-26
Fancyboost

### LINKED ISSUE #245: Fancypic?
Ignore fancypic in lgr, and have fancypic as a settings toggle(?)

GauChoob:
Resolves https://github.com/elmadev/elma-classic/issues/245

### DIFF
--- a/src/pic/lgr.cpp
+++ b/src/pic/lgr.cpp
@@ -33,6 +33,40 @@ bike_box BikeBox2 = {32, 183, 147, 297};
 bike_box BikeBox3 = {146, 141, 273, 264};
 bike_box BikeBox4 = {272, 181, 353, 244};
 
+constexpr char FANCYBOOST_NAMES[256][10] = {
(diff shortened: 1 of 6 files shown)


src/pic/lgr.cpp
```diff
@@ -638,6 +702,13 @@ lgrfile::lgrfile(const char* lgrname) {
             ERROR_CORRUPT();
         }
 
+        // Skip fancyboost-pattern files as we will internally construct them
```
jgouly, line 705: Only skip if fancyboost is enabled? Otherwise an LGR that is fancyboosted, but the option disabled, is un-boosted

GauChoob: That was my intention. Instead of the player tracking down and trying to keep track of which lgr files have been fancyboosted or not, just handle it all via an option internally

So you can have a fancyboosted file (that you downloaded the lgr from elma.online) but you don't want fancyboost, you can still disable it in options

Ben3eeE (approved):

PR #633, opened 2026-08-05
Fps/Ups Part 2

GauChoob:


include/physics/pacer.h
```diff
@@ -0,0 +1,28 @@
+#ifndef EOL_PACER_H
+#define EOL_PACER_H
+
+#include "main.h"
+
+namespace pacer {
+
+constexpr double STOPWATCH_TO_PHYS_TIME = 0.0024;
```
Ben3eeE: This magic number is used in other places (recorder, timer, pic), not just the ones changed here. Consider refactoring to use this constexpr everywhere instead of a magic number.

src/renderer/render.cpp
```diff
@@ -828,24 +829,63 @@ static void render_view(bool player1, bool bottom_player, pic8* pic, double time
     std::vector<info_panel_row> info_rows;
 
     if (bottom_player && loop != GameLoop::Render) {
+        // Limiter
+        auto format_limiter = [](std::string& text, bool fps) {
```
Ben3eeE: This kinda makes sense to have as a function in pacer instead of inline in `render_view`.

src/physics/pacer.cpp
```diff
@@ -0,0 +1,89 @@
...
+        if (enabled_changed) {
+            StatusMessages->add("Turning FPS limiter off when the next run starts");
+        } else {
+            StatusMessages->add("FPS limiter is already off");
+        }
+    } else {
+        bool limit_changed = limit != FpsLimit || limit != EolSettings->fps_limit();
+        if (enabled_changed || limit_changed) {
+            StatusMessages->add(
+                std::format("Setting FPS limiter to {} when the next run starts", limit));
```
Ben3eeE, line 59: This doesn't match the current behavior of eol. These are only added if show fps info is off. If show fps info is off the info is in the bottom right corner without a status message.

src/menu/options.cpp
```diff
@@ -444,6 +444,29 @@ void menu_options() {
         BOOL_OPTION("Show FPS:", show_fps);
         BOOL_OPTION("Show UPS:", show_ups);
 
+        nav.add_row(
+            "FPS Limit:",
+            EolSettings->fps_limit_enabled_persisted()
+                ? std::to_string(EolSettings->fps_limit_persisted())
+                : std::string("Off"),
+            NAV_FUNC() {
+                static constexpr int steps[] = {60, 100, 144, 240, 500};
```
Ben3eeE, line 454: This is really unusable 😅 but it's the only way that makes sense until we have a way to write exact number in settings. Same problem as number of chat lines.

Ben3eeE (commented):
Just took a quick look and have some initial comments. Want to take another deeper look.

src/physics/pacer.cpp
GauChoob: Imo it makes more sense to always show it. The message behaviour is also simplified a bit from eol but imo this makes sense. So it's intentionally a bit different from EOL - if you guys are strongly against we can adjust it

wontfix

src/menu/options.cpp
GauChoob: Yeah settings improvement is on the eventual todo list but out of scope for this pr

wontfix in this pr

Ben3eeE:
`fps::count_fps()` counts skipped frames so if you limit to 60 it shows `2000 (60)` which contradicts eol and is very confusing. The first if your current fps and the latter is the limiter value. Now it looks like it's not working.

src/eol/console.cpp
```diff
@@ -92,6 +93,31 @@ void console::register_console_commands() {
...
+            return;
+        }
+        std::string s(text);
+        if (strcmpi(s.c_str(), "on") == 0) {
+            pacer::request_fps_limit(true, EolSettings->fps_limit());
+        } else if (strcmpi(s.c_str(), "off") == 0) {
+            pacer::request_fps_limit(false, EolSettings->fps_limit());
+        } else {
+            char* end = nullptr;
+            int fps = (int)std::strtol(s.c_str(), &end, 0);
```
Ben3eeE: Without a base (passing 0) we accept hex and octal values for example; is it intentional that we allow `!fps 0x64` for 100 fps or `!fps 0100` for 64 fps?

src/physics/pacer.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace pacer {
+
+bool FpsLimitEnabled = false;
+int FpsLimit = 0;
+
+namespace {
+
+// In physics units of time
+static double time = 0.0;
```
Ben3eeE: Remove static from all members in this anonymous namespace

src/renderer/render.cpp
```diff
@@ -828,24 +829,63 @@ static void render_view(bool player1, bool bottom_player, pic8* pic, double time
...
+        };
+
+        bool show_fps = EolSettings->show_fps();
+        bool show_ups = EolSettings->show_ups() && loop == GameLoop::Game &&
+                        current_camera.mode == CameraMode::Normal;
+
         // FPS
-        if (EolSettings->show_fps()) {
+        if (show_fps) {
             const double fps_value = fps::fps();
```
Ben3eeE: `fps::fps()` reads really badly. What about `fps::current_fps` and `fps::current_ups` or whatever the value means.

src/eol/console.cpp
GauChoob: I guess it's better to disable. I can't remember because it's been so long since I made this PR but I think at the time I figured that if someone really wanted to make it hexadecimal they could. But there's really no use case. So I will disable

src/physics/pacer.cpp
GauChoob: I forgot to do changelog, so added changelog for all the differences, including the updated message text here

src/renderer/render.cpp
GauChoob: Renamed to fps::format_ups() and fps::format_fps() and also handle the formatting internally

PR #634, opened 2026-08-05
fix(platform_sdl): query display resolution for desktop fullscreen

Ben3eeE:
On X11 the fullscreen transition is asynchronous, so querying the window size right after SDL_SetWindowFullscreen can return a stale pre-fullscreen size. The mismatch triggered an unexpected resolution change while the window was still being created. A desktop-fullscreen window always ends up at the desktop resolution, so query the display mode instead of the window size.


Ben3eeE:
another fun fact is that update_resolution calls persist on the width/height settings but it's never written to in this case unless you go into the settings menu which you can't because you are stuck crashing on startup

PR #638, opened 2026-08-08
fix(platform_sdl): keep on_resolution_change out of startup

Ben3eeE:
This is the 2nd commit from speffix branch. Opening as a separate PR because this is great. The first I still not sure about.


jgouly:
segfaults with macOS with notch.

GauChoob (commented):
Seems to work fine on windows both before and after the fix, so no strong opinion but code makes sense to me

Imo would be nice to add an ELMA_ASSERT that balls is initialized instead of segfaulting to catch future problems more gracefully

Ben3eeE:
Rebased to a different idea: Completely avoid `on_resolution_change` during startup. Should fix both macos issue and the X11 issue Spef had.

Also added assert for balls initialized.

GauChoob:
Code overall makes sense, but seems weird to conditionally skip on_resolution_change(). Feels like it would be simpler to have an is_initialized var to avoid having to handle the boolean return value when modifying the fullscreen status (window code is already heavy and complicated and this is yet another thing to keep track of/handle properly in multiple places, instead of just skipping on_resolution_change() in one precise situation)

Ben3eeE:
This is removing the recursion that can happen by the platform doing these things which is a very confusing case. If we do anything like suggested we still have that, thought I admit I can't see the full shape of it because I don't understand when we are considered initialized. In this case it was Balls but could it be something else that causes issues in the future?

I argue this simplifies things more than anything else because of the removed recursion case.

jgouly (approved):

PR #639, opened 2026-08-09
Console commands

Ben3eeE:


src/eol/console.cpp
```diff
@@ -49,6 +50,15 @@ static std::optional<bool> parse_bool(std::string_view text) {
     return {};
 }
 
+static std::optional<int> parse_int(std::string_view text) {
+    int value = 0;
+    auto [ptr, ec] = std::from_chars(text.data(), text.data() + text.size(), value);
+    if (ec != std::errc() || ptr != text.data() + text.size()) {
+        return {};
```
GauChoob: Maybe nullopt? Since that's what we use elsewhere

src/eol/console.cpp
```diff
@@ -156,6 +160,42 @@ void console::register_console_commands() {
...
+                break;
+            default:
+                break;
+            }
+        }
+        const char* filename = current_level_filename();
+        add_line(std::format("name: {}", *filename ? filename : "(unnamed)"), LineType::Info);
+        add_line(std::format("title: {}", Level->level_name), LineType::Info);
+        add_line(
+            std::format("lgr: {}{}", Level->lgr_name, Level->lgr_not_found ? " (not found)" : ""),
```
GauChoob: Not sure lgr_not_found is a great variable to use. I almost deleted it a long time ago. But currently the lgr_name gets erased and replaced with "default" and lgr_not_found gets set to true.

As part of the cleanup I'm trying to prevent rendering from editing the level object (that means no deleting missing default ground/sky textures, no deleting missing pictures and no deleting missing lgr). At the moment I only did the second out of these three things.

tldr I prefer we keep it simple and `std::format("lgr: {}", Level->lgr_name)`

src/eol/console.cpp
```diff
@@ -156,6 +160,42 @@ void console::register_console_commands() {
     register_command("download_battle",
                      [](std::string_view /*text*/) { EolClient->download_battle_level(); });
     register_alias("dlb", "download_battle");
+    register_command("lev", [this](std::string_view) {
```
GauChoob: We might want to use !lev for something else eventually but I guess the name is ok until we have a better use for it(?)

GauChoob (approved):

src/eol/console.cpp
```diff
@@ -125,7 +159,43 @@ void console::register_console_commands() {
     register_alias("dl", "download");
     register_command("download_battle",
                      [](std::string_view /*text*/) { EolClient->download_battle_level(); });
-    register_alias("dlb", "download");
+    register_alias("dlb", "download_battle");
+    register_command("lev", [this](std::string_view) {
```
jgouly: Probably worth making this a static function rather than a lambda

PR #640, opened 2026-08-10
Flagtag battle

Ben3eeE:
Adds support for flagtag battles


src/eol/battle.cpp
```diff
@@ -134,6 +134,13 @@ bool eol::in_apple_battle() const {
            proto.playing_battle_level();
 }
 
+bool eol::kuski_has_flag(unsigned int kuski_id) const {
```
GauChoob, line 137: I'm wondering whether we should check if playing_battle_level() in this function but maybe it's an overkill check

GauChoob (approved):

PR #641, opened 2026-08-10
Fps limiter

Ben3eeE:
This is the FPS limiter from #633 with the non-eol additions removed, so we can merge a working, eol-faithful baseline first and discuss extensions (GPU display, message changes, UPS boost) separately afterwards. sunl's commits are preserved with original authorship.

Removed compared to #633:
- GPU display (show_gpu, !gpu, the GPU/UPS row); can be its own PR later
- Both changelog entries (one described features not in this PR, the other didn't match what eol actually does)

Changed compared to #633:
- Restores the 1e-6 first frame
- Restores eol's 33ms limiter timeout
- FPS in the bottom right corner shows the same value as eol; with the limiter off it can go past 1000
- Status messages behave like eol: only shown while the FPS display is hidden, since the display itself shows pending changes as (31) / (31 -> 60). eol was never silent: every `Log()` in its fps path is guarded by !ShowFPSInfo (`LimitFPS()`/`PutFPSChangePending()`), feedback just moves between the two surfaces. Not byte-identical (repeated identical !fps still re-prints where eol stays quiet), but close enough for the baseline.
- Sticky one-frame brake: this is part of eol's `LimitFPS()` itself (`ApplyOneFrameBrake` runs only on real frames); without it most one-frame-brake presses are dropped while the limiter is active
- Rewrote git history to be cleaner
- Also a theoretical issue fixed: `pacer::reset` ran before `enter_level` in #633, so lag during level entry inflated the first frame's elapsed time compared to main. It now runs where main's `stopwatch_reset` was.

Closes #633

### DIFF
--- a/src/eol/console.cpp
+++ b/src/eol/console.cpp
@@ -7,6 +7,7 @@
 #include "level/level.h"
 #include "level/object.h"
 #include "log.h"
+#include "physics/pacer.h"
 #include "pic/abc8.h"
 #include "platform/implementation.h"
 #include "platform/scancode.h"
@@ -167,6 +168,30 @@ void console::register_console_commands() {
     REGISTER_SETTINGS_INT(chat_lines);
 
     REGISTER_SETTINGS_BOOL(show_fps);
+
+    // eol-client !fps aggregate:
+    //   fps         -> toggle show_fps
+    //   fps on/off  -> queue limit on/off
+    //   fps <N>     -> queue numeric limit on
+    register_command("fps", [](std::string_view text) {
+        if (text.empty()) {
+            EolSettings->set_show_fps(!EolSettings->show_fps());
+            StatusMessages->add(EolSettings->show_fps() ? "FPS info shown" : "FPS info hidden");
+            return;
+        }
+        std::string s(text);
+        if (strcmpi(s.c_str(), "on") == 0) {
+            pacer::request_fps_limit(true, EolSettings->fps_limit());
+        } else if (strcmpi(s.c_str(), "off") == 0) {
+            pacer::request_fps_limit(false, EolSettings->fps_limit());
+        } else if (auto fps = parse_int(text); fps && *fps >= 30 && *fps <= 1000) {
+            pacer::request_fps_limit(true, *fps);
+        } else {
+            StatusMessages->add(
+                std::format("Invalid FPS: {}. Valid settings are 'on', 'off' or 30-1000", s));
+        }
+    });
+
     REGISTER_SETTINGS_BOOL(show_ups);
     register_alias("ups", "show_ups");
 
--- a/src/game/fps.cpp
+++ b/src/game/fps.cpp
@@ -36,7 +36,7 @@ void reset() {
     prev_time = get_milliseconds();
 }
 
-static void calculate() {
+void update() {
     long long update_interval = UPDATE_INTERVAL;
     if (prev_fps == 0.0) {
         update_interval = FIRST_UPDATE_INTERVAL;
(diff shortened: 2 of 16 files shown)


src/eol/console.cpp
```diff
@@ -92,6 +103,29 @@ void console::register_console_commands() {
...
+        std::string s(text);
+        if (strcmpi(s.c_str(), "on") == 0) {
+            pacer::request_fps_limit(true, EolSettings->fps_limit());
+        } else if (strcmpi(s.c_str(), "off") == 0) {
+            pacer::request_fps_limit(false, EolSettings->fps_limit());
+        } else if (auto fps = parse_int(text)) {
+            if (*fps >= 30 && *fps <= 1000) {
+                pacer::request_fps_limit(true, *fps);
+            }
+        }
```
GauChoob, line 192: Should probably have a fail condition message here ("invalid input: {}"), which should also catch fps values < 30 and > 1000

src/physics/pacer.cpp
```diff
@@ -43,11 +53,36 @@ void request_fps_limit(bool enabled, int limit) {
...
+    long long now = get_milliseconds();
+    long long elapsed = now - start_time;
+
+    if (FpsLimitEnabled) {
+        // In milliunits (a value of 1000 corresponds to one allowed frame)
+        long long max_allowed_frames = elapsed * FpsLimit;
+
+        // eol-client forces a real frame at least every 33 ms even when over the limit
+        if (real_frame_count * 1000LL > max_allowed_frames &&
+            now - last_real_frame_time <= LIMITER_TIMEOUT_MS) {
```
GauChoob, line 95: Imo we should drop last_real_frame_time and LIMITER_TIMEOUT_MS as per internal discussion

src/game/fps.cpp
```diff
@@ -46,12 +54,10 @@ static void calculate() {
...
-void count_fps() {
-    frame_count++;
-
-    calculate();
-}
+void count_fps() { frame_count++; }
 
 void count_ups() { update_count++; }
 
+void update() { calculate(); }
```
GauChoob: Can maybe just rename the calculate() function to update() instead of needlessly nesting functions here. We can always break it up in the future if ever

GauChoob (dismissed):
Yeah this is a good PR, imo we can merge and discuss the rest after.

Imo I would drop the 33 ms limiter thing as it's unneeded and the "wrong" fix to a resolved problem as discussed in private

Please drop this commit for this PR (we can submit the commit in its own PR maybe after) [fix(game): make one-frame brake sticky across outer iters](https://github.com/elmadev/elma-classic/pull/641/changes/d7996e2818a0efedeaa0a0ded7176be383cbcfc0)

jgouly:
I think we should keep the timeout etc, to try have a baseline that matches, code wise too, what EOL currently does. We can drop it with reasonings + testing etc.

GauChoob:
Link to LIMITER_TIMEOUT test showing it does nothing: https://github.com/elmadev/elma-classic/compare/fps_limiter_test?expand=1

src/eol/console.cpp
GauChoob: I was thinking something ish along the lines of

```cpp
        if (strcmpi(s.c_str(), "on") == 0) {
            pacer::request_fps_limit(true, EolSettings->fps_limit());
        } else if (strcmpi(s.c_str(), "off") == 0) {
            pacer::request_fps_limit(false, EolSettings->fps_limit());
        } else if (auto fps = parse_int(text) && *fps >= 30 && *fps <= 1000) {
            pacer::request_fps_limit(true, *fps);
        } else {
            StatusMessages->add(std::format("Invalid FPS: {}. Valid settings are 'on', 'off' or 30-1000", s.c_str()));
        }
```

GauChoob (approved):

PR #645, opened 2026-08-12
Organize changelog

GauChoob:
console commands probably need to be prepended to specific lines or otherwise have their own section - this is not addressed in this commit

I improved the text, sorted the entries into categories and sort-of-ish sorted each category in order of importance (kind of subjective)

It's not necessarily the final order/categorization but it makes the changelog a lot more manageable now

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,116 +1,147 @@
-* Version 1.4
+# Version 1.4
 
 ## Major Features
-* Polygons no longer have a per-polygon max vertex count. There is now only a maximum total vertex count for the entire level.
-* Steam elma.res is supported
-* Elma now honors the local keyboard layout for text input instead of being forced to US qwerty layout
-* QGRASS texture is available in internal editor
-* Replays: maximum duration increased from 1 hour to about 828 days.
-* Add support for LGR13
-* Add setting to properly scale grass with zoom
-* There is now navigation to "Level replays" after playing a level to view replays in that level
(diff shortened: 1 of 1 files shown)


Ben3eeE:
58 [organization] 'About menu updated' is filed under the '### Key Input' section where it does not belong
129 [wording]      'Missing pixels were added at some connection points between two polygon lines' reads as describing a bug, not a fix
139 [typo]         Stray trailing comma: '* Disabled noretrac.inf (software backbuffer),'
1 [formatting]   File now begins with a stray blank line before '# Version 1.4'
37 [completeness] Rewrite drops the 'can also set it via keybind' half of the force-texture entry
81 [completeness] Rewrite drops the 'or wrong lgr used' condition from the texture-reset fix

GauChoob:
Thanks benai

```
58 [organization] 'About menu updated' in wrong category ✅
129 [wording] 'Missing pixels were added' reads as describing a bug, not a fix ✅ passive -> active tense
139 [typo] Stray trailing comma: 'noretrac.inf,' ✅
1 [formatting] File now begins with a stray blank line before '# Version 1.4' ✅ deleted line for you
37 [completeness] Rewrite drops the 'can also set it via keybind' half of the force-texture entry ✅ Console and Shortcuts section created
81 [completeness] Rewrite drops the 'or wrong lgr used' condition from the texture-reset fix ❌ Intentional, does not need this information
```

Ben3eeE (approved):

PR #647, opened 2026-08-12
Speedometer

### LINKED ISSUE #510: Spedometer

GauChoob:
Resolves https://github.com/elmadev/elma-classic/issues/510

### DIFF
--- a/include/game/state.h
+++ b/include/game/state.h
@@ -104,6 +104,7 @@ class state {
     combo_scancode key_show_chat;
     combo_scancode key_battle_status;
     combo_scancode key_battle_leader;
+    combo_scancode key_speedometer;
     combo_scancode key_reconnect;
     combo_scancode key_disconnect;
     combo_scancode key_toggle_one_wheel_status;
--- a/src/eol/settings.cpp
+++ b/src/eol/settings.cpp
@@ -190,6 +190,7 @@ void eol_settings::set_chat_key(combo_scancode key) { chat_key_ = key; }
 void eol_settings::set_show_chat_key(combo_scancode key) { show_chat_key_ = key; }
 void eol_settings::set_battle_status_key(combo_scancode key) { battle_status_key_ = key; }
 void eol_settings::set_battle_leader_key(combo_scancode key) { battle_leader_key_ = key; }
+void eol_settings::set_speedometer_key(combo_scancode key) { speedometer_key_ = key; }
 void eol_settings::set_reconnect_key(combo_scancode key) { reconnect_key_ = key; }
 void eol_settings::set_disconnect_key(combo_scancode key) { disconnect_key_ = key; }
 void eol_settings::set_toggle_one_wheel_status_key(combo_scancode key) {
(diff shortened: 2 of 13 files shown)


src/eol/settings.cpp
```diff
@@ -575,7 +579,7 @@ void eol_settings::sync_controls_from_state(state* s) {
     EolSettings->persist_battle_results_key(s->key_battle_results);
     EolSettings->persist_chat_key(s->key_chat);
     EolSettings->persist_battle_status_key(s->key_battle_status);
-    EolSettings->persist_battle_leader_key(s->key_battle_leader);
+    EolSettings->persist_speedometer_key(s->key_speedometer);
```
Ben3eeE, line 685: `persist_battle_leader_key` disappeared

src/menu/options.cpp
```diff
@@ -477,7 +477,7 @@ void menu_options() {
 
         BOOL_OPTION("Show others:", show_others);
         BOOL_OPTION("Show battle status:", show_battle_status);
-        BOOL_OPTION("Show battle leader:", show_battle_leader);
+        BOOL_OPTION("Show speedometer:", show_speedometer);
```
Ben3eeE, line 533: `Show battle leader` disappeared

include/game/state.h
```diff
@@ -98,6 +98,7 @@ class state {
     combo_scancode key_chat;
     combo_scancode key_battle_status;
     combo_scancode key_battle_leader;
+    combo_scancode key_speedometer;
```
Ben3eeE, line 107: There is no code added to actually use this keybinding.

src/renderer/render.cpp
```diff
@@ -827,6 +827,11 @@ static void render_view(bool player1, bool bottom_player, pic8* pic, double time
     // rows are rendered in the order they were added (last added on top)
     std::vector<info_panel_row> info_rows;
 
+    if (EolSettings->show_speedometer() && current_camera.mode != CameraMode::MapViewer) {
+        info_rows.push_back({"max speed", std::move(driv.speed.format_max_speed())});
```
Ben3eeE: These `std::move` are redundant imo. Joey can maybe confirm.

Ben3eeE (dismissed):
Do we really need a separate class for speedometer? It's not gonna grow more complicated, we need other metadata as well. What do we plan on giving `exit_level`? The whole driver object? Should there be a metadata class instead or should it just be a struct on driv?
Doesn't use existing vec2 len function and instead makes it's own calculation.

Should speedometer be reset on flag tag respawn that it would be if it was in `reset_metadata`?

Commit messages are lazy and could use a few more words.

GauChoob:
```
Do we really need a separate class for speedometer? It's not gonna grow more complicated, we need other metadata as well.
```
Renamed to motor_stats struct within driver.cpp

```
What do we plan on giving exit_level? The whole driver object? Should there be a metadata class instead or should it just be a struct on driv?
```
Yes we should pass entire driver. I changed from class to struct with public access to vars

```
Doesn't use existing vec2 len function and instead makes it's own calculation.
```
Fixed, I was copying old code and forgot about len function

```
Should speedometer be reset on flag tag respawn that it would be if it was in reset_metadata?
```
No

```
Commit messages are lazy and could use a few more words.
```
Improved

GauChoob:
Currently !speedometer displays different text from using the Ctrl+F10 shortcut (show_speedometer: on/off vs speedometer shown/hidden)

It's an easy fix but we can either leave as is for now or come up with a specific solution if someone has a strong opinion about how to go about it

Ben3eeE (approved):

PR #652, opened 2026-08-15
Add show chat setting

Ben3eeE:


jgouly (approved):

GauChoob (approved):
Unteste dbut makes sense to me

PR #654, opened 2026-08-15
feat(battle): show battle attributes in battle results title

Ben3eeE:
These are completely hidden in eol currently imo so this is a proposal to show the information in f6 table title.

### DIFF
--- a/src/eol/battle.cpp
+++ b/src/eol/battle.cpp
@@ -119,13 +119,38 @@ std::string eol::format_level(std::string_view level) {
     return with_ext;
 }
 
+// First Finish battle results in Sju250.lev (apple bugs, see others)
 void eol::set_battle_results_title(const char* label) {
+    using namespace BattleAttributes;
+
+    constexpr std::pair<Kind, std::string_view> extras[] = {
+        {AcceptBugs, "apple bugs"},
+        {AllowStarter, "allow starter"},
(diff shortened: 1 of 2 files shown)


src/eol/battle.cpp
```diff
@@ -120,12 +120,39 @@ std::string eol::format_level(std::string_view level) {
...
+        add_detail(std::format("{} apple{}", apple_count, apple_count == 1 ? "" : "s"));
+    }
+    for (auto [flag, text] : extras) {
+        if (current_battle->attributes & flag) {
+            add_detail(text);
+        }
     }
+
+    if (!details.empty()) {
+        new_title += std::format(" ({})", details);
```
GauChoob, line 151: Is it more efficient and/or better practice to do 

`new_title = std::format("{} ({})", new_title, details);`

src/eol/battle.cpp
```diff
@@ -120,12 +120,39 @@ std::string eol::format_level(std::string_view level) {
 }
 
 void eol::set_battle_results_title(const char* label) {
```
GauChoob, line 123: Can we add a comment somewhere here just to see what the string looks like

e.g.

// Apple battle results in LEV.lev (1 apple, see others)

GauChoob (approved):
Imo I agree with adding these battle attributes

PR #656, opened 2026-08-15
fix one frame brake

Ben3eeE:


src/game/game.cpp
```diff
@@ -666,6 +679,13 @@ int game_loop(const char* filename, CameraMode camera_mode) {
                 time += dt;
             }
 
+            // Input is fixed for a whole physics frame; a press on a frame
+            // that ran no subframes stays pending until physics advances
+            if (ran_subframes) {
```
GauChoob, line 695: We could possibly just check if(dt > 0.0) here instead of creating ran_subframes variable

src/game/game.cpp
```diff
@@ -593,8 +599,15 @@ int game_loop(const char* filename, CameraMode camera_mode) {
...
 
+            latch_one_frame_brake(driv1);
+            if (!Single) {
+                latch_one_frame_brake(driv2);
+            }
+
+            bool ran_subframes = false;
             double dt = 0.0;
             while (pacer::subframe(&dt)) {
+                ran_subframes = false;
```
GauChoob: if we keep ran_subframes, then this should be ran_subframes = true;

GauChoob (changes requested):

src/renderer/render.cpp
```diff
@@ -852,6 +853,10 @@ static void render_view(bool player1, bool bottom_player, pic8* pic, double time
         }
     }
 
+    if (loop == GameLoop::Replay) {
+        info_rows.push_back({"Level", current_level_filename()});
```
GauChoob: Delete this debug code

src/game/game.cpp
GauChoob: I still prefer reusing dt and checking if(dt > 0.0) as opposed to ran_subframes

GauChoob (approved):
I leave to joey to decide if he prefers dt > 0.0 or ran_subframes. I prefer dt > 0.0

jgouly:
I felt the bool was more clear

PR #657, opened 2026-08-15
feat(renderer): show spied kuski run time on timer

Ben3eeE:


GauChoob (approved):
Strange order when spy data is not yet interpolated but I guess it's ok for now

jgouly:
I will give this a test n merge

PR #665, opened 2026-08-16
Fps booster

GauChoob:
Play at 1000 fixed UPS even with a crappy computer

Even with a screen refresh rate of 1000-1400 you often drop frames, or if your computer has background processes you can drop frames every once in a while, which can be a problem for Uphill Battle. This guarantees 1000 UPS no matter what.

1) I kept FPSLimitEnabled and FPSBoost as separate bools instead of an enum. If we prefer, we could merge them together
2) It's unclear what FPS should display with the booster since currently it's currently showing either frame rate or throttled outer loop rate. Easiest thing would be to just display the UPS if the FPSBoost is on.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -25,6 +25,7 @@
 ### Gameplay
 * Internal level total time is displayed in menu
 * All internal levels may be immediately unlocked via the Options
+* Game FPS can be boosted to the max of 1000, regardless of the computer's processing power and screen display settings
 * When saving an internal replay, the internal number and name will be shown instead of QWQUU0XX.lev
 * !levinfo prints basic information about the current level
 * Save play is enabled for internal level 55 (More Levels)
--- a/include/eol/settings.h
+++ b/include/eol/settings.h
@@ -147,6 +147,7 @@ class eol_settings {
     Default<bool> show_ups_{false};
     Default<bool> fps_limit_enabled_{false};
     Clamp<int> fps_limit_{30, 100, 1000};
+    Default<bool> fps_boost_{false};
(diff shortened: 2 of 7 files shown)


Ben3eeE:
I'm against this change because it creates situations that are impossible on any hardware.

Input is still sampled at render rate which means that if you are 60 fps you will have about 16ms input lag, and if you are 500 fps you have 2ms input lag. It's basically like vsync but in reverse which makes it less useful and not something that any hardware can produce today.

Also 1fb changes meaning from braking for one frame to braking for one render frame which means many more physics frames. Another new situation.

Ultimately I guess it's not just my decision what is done but this needs to be tested and agreed with others outside the development team before implemented and released to everyone which means it's something I am also against merging before we have released something to people so there can be a new version where this is tested and decided upon. Which is what has been done for other changes in eol (Fixing first frame fps limiter bug, adding 1fb etc).

GauChoob:
I guess I can agree with having this tested as a prototype instead of merging now. Imo this is not disruptive to the playing paradigm, unlike fps limiter and one-frame-brake which both completely changed the WR progression progress.

Having delayed/slower inputs is only a slight downside but not really abuseable. Only abuseable part is artificially limiting your Hz to 60 and trying to play at boosted FPS but imo I don't really see how to take advantage of it.

And it's incredibly annoying having to adjust zoom/quality/screen resolution for Uphill Battle to get a resemblance of 1000 fps (when in reality you're probably only achieving 980 UPS or something).

tldr there's a few small details to figure out eventually but we should definitely trial this on one of the releases

PR #669, opened 2026-08-18
feat(console): add !persist command

Ben3eeE:


GauChoob:
Imo persist is a good internal name but not a good public name. Maybe !save or something instead?

Ben3eeE:
!save is ambiguous with replays/levels/state.dat and undersells the actual semantics - this command doesn't save something unsaved, it promotes intentionally session-only values to persisted ones, which is exactly what the internal persisted concept is called.

GauChoob:
Still needs a better name. I've never seen "persist" as a term in any game's settings. !save conf or !save_conf would be better to me

Spef-dat:
I think this should exist as a more verbose setting in the Options menu, which most users would use. Something like "Remember in-level changes". Then the name of the console command matters less and !persist would be ok for advanced use.

PR #675, opened 2026-08-20
Interpol spies

### LINKED ISSUE #543: Spy bugs
1. ~~f1 to other players hidden when spying someone hides the player you are spying in addition to everyone else, but the camera still moves~~
2. ~~f2 should only write "not observing anyone anymore" if you were observing someone (go to int55 and spam f2 to enjoy status message spam)~~
3. It doesn't remove player when they die, only resets once they enter level again, eol resets camera back to start when kuski dies and removes them from view.

Ben3eeE:
Fixes https://github.com/elmadev/elma-classic/issues/543


include/eol/kuski.h
```diff
@@ -22,9 +23,21 @@ class spy_playback {
   public:
     const struct spy_data* spy_data() const;
     void add(const struct spy_data& sd);
+    void stop();
```
GauChoob, line 27: Can we clarify distinction between stop and clear and rebuffer?

stop = clear buffered frame data and hide kuski
rebuffer = clear buffered frame data and freeze kuski(?)
clear = clear buffered frame data without freezing kuski(?)

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+        d1.mot.flipped_bike == d2.mot.flipped_bike);
+
+    return out;
+}
+
+// Returns the position just after the newest non-stop frame at or before
+// time, or frames.begin() if there is none. frame_time maps a frame onto
+// whatever clock time is on.
+template <typename Frames, typename FrameTime>
+static auto position_after_newest_at_or_before(Frames& frames, uint32_t time,
```
GauChoob: I still think this is more complicated than it needs to be.

Imo at the most basic level, the intent of the function is:

// returns the next frame to occur after the specified time
static auto next_frame(...) {
   ...
   // stop frames don't have a valid time property
}

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+    // just show the newest pose with the run clock at zero.
+    if (!frames.empty() && !frames.back().stop && frames.back().data.time == 0) {
+        data = frames.back().data;
+        frames.erase(frames.begin(), frames.end() - 1);
+        return;
+    }
+
+    if (running) {
+        // Advance playback: everything before the frame now falls on is spent.
+        auto pos = position_after_newest_at_or_before(
```
GauChoob: Possible to do 

```cpp
        auto pos = position_after_newest_at_or_before(
            frames, mask24(now - time_offset), [&](const spy_frame& f) { return f.data.time; });
```

?

If this is possible, then we can get rid of the templated FrameTime

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+
+    return out;
+}
+
+// Returns the position just after the newest non-stop frame at or before
+// time, or frames.begin() if there is none. frame_time maps a frame onto
+// whatever clock time is on.
+template <typename Frames, typename FrameTime>
+static auto position_after_newest_at_or_before(Frames& frames, uint32_t time,
+                                               FrameTime frame_time) {
```
GauChoob: I think we can replace typename Frames with

`std::deque<spy_playback::spy_frame>& frames`

GauChoob: Therefore both templated parameters don't need to be templated anymore, greatly simplifying the function?

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+            frames, now, [&](const spy_frame& f) { return mask24(f.data.time + time_offset); });
+        if (pos == frames.begin()) {
+            // Even the oldest frame is ahead of the playback clock.
+            // This should never happen under normal situations; it takes
+            // a stall long enough (hours) to flip the wrapping compares.
+            clear();
+            return;
+        }
+        frames.erase(frames.begin(), pos - 1);
+    } else if (min_spy_frames > 0 && frames.size() >= static_cast<size_t>(min_spy_frames)) {
```
GauChoob, line 213: Does the server still send spy frames if min_spy_frames == 0 or if it is negative?

We should either delete this check, or separate this check into its own conditional since it's not directly related to this clause

src/eol/kuski.cpp
GauChoob: From what I understand, as soon as you get a new run, the frames data is cleared
- add() -> rebuffer if run id changes
- update() -> clear if stop encountered

So frames always contains data about a single run
Therefore it should be fine to search both forwards and backwards
Given that we call the function once in add() to get what is likely the last frame, and we call the function once in update() to get what is close to the first frame, maybe it's better to search forwards in the list as it is easier to follow and codeflow-wise makes more sense to me

Presumably the function originally searched backwards because you were calling add() before update(), but forward search seems more natural

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+    }
+
+    const spy_frame& frame1 = frames.front();
+    const spy_frame& frame2 = frames.size() > 1 ? frames[1] : frame1;
+
+    // Playback hit a stop marker, or frame2 is so old that fresh frames
+    // should long since have arrived. The run is either over or the feed
+    // died; hide the kuski.
+    constexpr uint32_t STALE_FRAME_MS = 3000;
+    if (frame1.stop || frame2.stop ||
```
GauChoob, line 230: Could be nice to consider here a slight change to behaviour

```cpp
spy_frame*frame1;
spy_frame*frame2;
if(frame2.stop) {
    frame2 = frame1;
}
if(frame1.stop) {
    clear();
}
```

Which keeps the bike alive in its "death" state for 1 frame instead of instantly disappearing

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+        compare_wrapping(now, mask24(frame2.data.time + time_offset + STALE_FRAME_MS), 24) == 1) {
+        clear();
+        return;
+    }
+
+    // unwrap values so they respect the order time1 -> time2 -> now
```
GauChoob: Usual order would be time1 -> now -> time2?

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+    const uint32_t time1 = mask24(frame1.data.time + time_offset);
+    uint32_t time2 = mask24(frame2.data.time + time_offset);
+    if (time2 < time1) {
+        time2 += 1 << 24;
+    }
+    if (now < time1) {
+        now += 1 << 24;
+    }
+
+    // alpha runs past 1 when the next frame is late; the clamp parks the pose
```
GauChoob: I'm surprised that the frames don't get purged before alpha > 1.0 but I guess I'd have to test this or you tested this

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
 }
 
+static uint32_t mask24(uint32_t val) { return val & 0xFFFFFF; }
+
+// val1 and val2 are monotonically increasing values that wrap around at 2^bits
+// compare these two values, handling the possibility of this wraparound
+static int compare_wrapping(uint32_t val1, uint32_t val2, int bits) {
+    const uint32_t max_diff = 1u << (bits - 1);
+
+    if ((val1 < val2 && val2 - val1 < max_diff) || (val1 > val2 && val1 - val2 >= max_diff)) {
```
GauChoob, line 24: Could be nice to have a comment like
// -1: val1 < val2
// 0: val1 = val2
// 1: val1 > val2

To quickly see what the "normal" interpretation of each returned value is

It's what you would expect but it's nice to be explicit

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+    }
+    if (angle2 - angle1 > M_PI) {
+        return (1 - alpha) * angle1 + alpha * (angle2 - 2 * M_PI);
+    }
+    return (1 - alpha) * angle1 + alpha * angle2;
+}
+
+// Interpolating wheel/head positions linearly makes them drift off a rotating
+// bike, so go through polar coordinates around the bike center instead.
+static vect2 interpolate_around_center(const vect2& p1, const vect2& p2, double alpha,
```
GauChoob, line 47: Currently codebase uses r1 and r2 as vector points

So it doesn't make sense to me for length to be r1 and r2

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+                                       const vect2& center1, const vect2& center2) {
+    const double r1 = (p1 - center1).length();
+    const double r2 = (p2 - center2).length();
+
+    double a1 = 0.0;
+    if (r1 > 0) {
```
GauChoob: 0.0 instead of 0

src/eol/kuski.cpp
```diff
@@ -1,13 +1,258 @@
...
+        a1 = acos((p1.x - center1.x) / r1);
+        if (p1.y - center1.y < 0) {
+            a1 = 2 * M_PI - a1;
+        }
+    }
+
+    double a2 = 0.0;
```
GauChoob, line 60: AI suggested to simplify with

```
    const double a1 = r1 > 0.0 ? atan2(d1.y, d1.x) : 0.0;
    const double a2 = r2 > 0.0 ? atan2(d2.y, d2.x) : 0.0;
```

But I didn't take the time to work through the trigonometry math to see if it is true or not

src/eol/kuski.cpp
jgouly: The server doesn't use the min spy frames value at all, just sends it to the client on setup.

src/eol/kuski.cpp
jgouly: spy_frame is private, it's just templated on Frames now, better than the 2 types.

src/eol/eol.cpp
```diff
@@ -375,6 +375,10 @@ void eol::download_battle_level() {
 }
 
 void eol::enter_level(const char* level_name, const level* lev, bool spying) {
+    for (kuski& k : kuskis_) {
```
jgouly: We do this on exit and enter now?

jgouly (approved):
Looks good and seems to match EOL (while the codes looks very different)

src/eol/kuski.cpp
```diff
@@ -1,13 +1,259 @@
...
+}
+
+// Only blend while the discrete state behind the phase (volt direction, bike
+// flip) matches; otherwise snap. Also keep the result off exactly 0.5, which
+// sits right on sign flips in render_bike (the wheel z-order, for one).
+static double interpolate_anim_phase(double alpha, double phase1, double phase2,
+                                     bool discrete_matches) {
+    double phase =
+        discrete_matches ? (1 - alpha) * phase1 + alpha * phase2 : (alpha < 0.5 ? phase1 : phase2);
+    if (fabs(phase - 0.5) < 1.0 / 256) {
```
GauChoob, line 81: Why do we want to keep the phase off exactly 0.5? I know we used to get an ASAN error/crash at exactly 0.5 but this is fixed now. In normal replay playback we don't have this guard

src/eol/kuski.cpp
```diff
@@ -1,13 +1,259 @@
...
+        alpha, d1.metadata.bike_turning.turn_phase, d2.metadata.bike_turning.turn_phase,
+        d1.mot.flipped_bike == d2.mot.flipped_bike);
+
+    return out;
+}
+
+// Returns the position just after the newest non-stop frame at or before
+// time, or frames.begin() if there is none. offset shifts each frame's
+// time onto the same clock as time.
+template <typename Frames>
```
GauChoob, line 123: Just make this a private class function instead of making it a template?

src/eol/kuski.cpp
```diff
@@ -1,13 +1,259 @@
...
+static auto position_after_newest_at_or_before(Frames& frames, uint32_t time, uint32_t offset) {
```
GauChoob, line 124: Sorry to say it but function name is still garbage

after_newest -> the `newest` variable only exists in spy_playback::add() and not spy_playback::update()
after_at_or_before -> it's pretty non-sensical to have all 3 signs (>, =, <) in the function name

here's a really simple function name:
```cpp
// return a pointer to the next frame occurring after the specified time
frame_after_time()
```

src/eol/kuski.cpp
```diff
@@ -1,13 +1,259 @@
...
+    auto pos = frames.end();
```
GauChoob, line 125: A stop frame is guaranteed to always be at the end of the queue because all added frames are always placed in front of it and also spy_playback::stop() places the stop frame at the end of the queue

If we assume the stop frame is always at the end, forward/backwards search is identical, so this would be a lot simpler to read:

```cpp

std::deque<spy_playback::spy_frame>::iterator spy_playback::frame_after_time(uint32_t time,
                                                                             uint32_t offset) {
    for (auto pos = frames.begin(); pos != frames.end(); ++pos) {
        if (pos->stop) {
            // end of the queue
            // stop frames have no defined time
            return pos;
        }
        if (compare_wrapping(time, mask24(pos->data.time + offset), 24) == -1) {
            return pos;
        }
    }
    return frames.end();
}
```

src/eol/kuski.cpp
```diff
@@ -1,13 +1,259 @@
...
+        // Advance playback: everything before the frame now falls on is spent.
+        auto pos = position_after_newest_at_or_before(frames, now, time_offset);
+        if (pos == frames.begin()) {
+            // Even the oldest frame is ahead of the playback clock.
+            // This should never happen under normal situations; it takes
+            // a stall long enough (hours) to flip the wrapping compares.
+            clear();
+            return;
+        }
+        frames.erase(frames.begin(), pos - 1);
```
GauChoob, line 212: Mathematically wouldn't this prevent alpha from being greater than 1.0 below?

position_after_newest_at_or_before() only allows one frame prior to now, so this implies that frame2 occurs after now (except when frame1 = frame2 and we don't interpolate)

src/eol/kuski.cpp
GauChoob: So we never expect min_spy_frames to <= 0 and we should block or error on this in spy_playback::add(), not here

GauChoob (changes requested):
Went through the PR again and here are remaining comments

As said before, looking good but still a few small things that I'm not sure about/unclear

src/eol/kuski.cpp
jgouly: Lets keep it, do some more testing without and remove it

src/eol/kuski.cpp
jgouly: From a purely pedantic standpoint going fowards means you, on average, are likely to iterate over more frames, because the frames likely come in the right order, so the frame is likely at the end (but there is likely to be very few frames)

src/eol/kuski.cpp
jgouly: Possibly but leaving the clamp for now, can come back to this

src/eol/kuski.cpp
GauChoob: I don't think it's worth optimizing for CPU, just for legibility and conceptually, but to be pedantic forward search actually saves more time:

add() iterates backwards (usually taking last frame), but is only called once every data_interval milliseconds
update() iterates forward (usually taking first frame) and is called every single frame

Therefore forward search is more efficient by a factor of data_interval

jgouly:
Fixed comments, other few comments will be for later testing / tweaking

GauChoob (approved):
Conditional on test

